PR #3, opened 2016-02-03
Add .travis.yml

gilles-duboscq:


mx.graal/mx_graal_8.py
```diff
@@ -248,74 +245,30 @@ def run(self, tasks, extraVMarguments=None):
...
 
 def compiler_gate_runner(suites, unit_test_runs, bootstrap_tests, tasks, extraVMarguments=None):
-
-    # Build server-hosted-jvmci now so we can run the unit tests
-    with Task('BuildHotSpotGraalHosted: product', tasks) as t:
-        if t: buildvms(['--vms', 'server', '--builds', 'product'])
-
     # Run unit tests on server-hosted-jvmci
-    with VM('server', 'product'):
+    with VM('jvmci', 'fastdebug'):
```
rschatz: This line is wrong... The graal unittests are designed in a way that they compare the default behavior of the VM with an explicit invocation of graal, so if we run the VM in jvmci mode, we're comparing graal with itself.

gilles-duboscq: That's true, we are "probably" testing against the interpreter though. Marking the gate minimal meant building a single VM flavor.

PR #16, opened 2016-02-10
remove redundant method substitution mechanism

dougxc:
Currently, method intrinsification by means of substituting the body of one method for another can be done by one of two ways. The first (and original) way involves using the `@MethodSubstitution` and `@ClassSubstitution` annotations and registering substitutions via `Replacements.registerSubstitutions()`. After `GraphBuilderPlugin`s were introduced, an alternative mechanism embodied by `MethodSubstitutionPlugin` was added. This change converts all existing uses of the old mechanism to the new mechanism and removes the former.

Note that `@MethodSubstitution` and `@ClassSubstitution` have been preserved (along with the annotation processor that validates their usages). However, they are mostly for informational purposes now (i.e. to clearly denote that a given method is a substitute).


dougxc:
Can you please review this as well @christhalinger..

graal/com.oracle.graal.api.replacements/src/com/oracle/graal/api/replacements/MethodSubstitution.java
```diff
@@ -30,8 +30,11 @@
 import jdk.vm.ci.meta.Signature;
 
 /**
- * Denotes a substitute method. A substitute method can call the original/substituted method by
- * making a recursive call to itself.
+ * Denotes a method whose body is used by a compiler as the substitute (or intrinsification) of
+ * another method. The exact method used to do the substitution is compiler dependent by every
```
christhalinger: "by" seems to be a typo.

christhalinger:
I've said this before but in my opinion this is a step backwards.  Can you explain once more why you are changing this?

At least we are keep the annotations which is very helpful.

dougxc:
The main reason for this change is to remove a redundant mechanism. I may be able to address your remaining concerns better if you explain why you think that is a step backwards.

christianwimmer:
@christhalinger Currently there are two mechanism to register method substitutions, which is unnecessary complexity.  The "new" mechanism allows substitutions to be performed during bytecode parsing.

christhalinger:
Let's take `ConstantPoolSubstitutions` as an example.  Before we had code like:

```
 -    private static Word metaspaceConstantPool(Object constantPoolOop) {
 -        // ConstantPool.constantPoolOop is in fact the holder class.
 -        Class<?> constantPoolHolder = (Class<?>) constantPoolOop;
 -        KlassPointer klass = ClassGetHubNode.readClass(constantPoolHolder);
 -        return klass.readWord(instanceKlassConstantsOffset(), INSTANCE_KLASS_CONSTANTS);
 -    }
```

and now:

```
 +    private static boolean readMetaspaceConstantPoolElement(GraphBuilderContext b, ValueNode constantPoolOop, ValueNode index, JavaKind elementKind, WordTypes wordTypes, HotSpotVMConfig config) {
 +        ValueNode constants = getMetaspaceConstantPool(b, constantPoolOop, wordTypes, config);
 +        int shift = CodeUtil.log2(wordTypes.getWordKind().getByteCount());
 +        ValueNode scaledIndex = b.add(new LeftShiftNode(index, b.add(ConstantNode.forInt(shift))));
 +        ValueNode offset = b.add(new AddNode(scaledIndex, b.add(ConstantNode.forInt(config.constantPoolSize))));
 +        AddressNode elementAddress = b.add(new OffsetAddressNode(constants, offset));
 +        boolean notCompressible = false;
 +        ValueNode elementValue = WordOperationPlugin.readOp(b, elementKind, elementAddress, NamedLocationIdentity.getArrayLocation(elementKind), BarrierType.NONE, notCompressible);
 +        b.addPush(elementKind, elementValue);
 +        return true;
 +    }
```

This is obviously more complex.  Having to add nodes to the graph manually is a step backwards in my opinion ("Hello, C2!").  Not to mention the actual replaced methods:

```
 -    @MethodSubstitution(isStatic = false)
 -    private static int getSize0(@SuppressWarnings("unused") final ConstantPool thisObj, Object constantPoolOop) {
 -        return metaspaceConstantPool(constantPoolOop).readInt(constantPoolLengthOffset());
 -    }
```

vs.

```
 +        r.register2("getSize0", Receiver.class, Object.class, new InvocationPlugin() {
 +            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode constantPoolOop) {
 +                boolean notCompressible = false;
 +                ValueNode constants = getMetaspaceConstantPool(b, constantPoolOop, wordTypes, config);
 +                AddressNode lengthAddress = b.add(new OffsetAddressNode(constants, b.add(ConstantNode.forLong(config.constantPoolLengthOffset))));
 +                ValueNode length = WordOperationPlugin.readOp(b, JavaKind.Int, lengthAddress, CONSTANT_POOL_LENGTH, BarrierType.NONE, notCompressible);
 +                b.addPush(JavaKind.Int, length);
 +                return true;
 +            }
 +        });
```

christhalinger:
@christianwimmer I understand but replacing a simple solution with something more complex makes maintainability go down.  What's the advantage of doing substitutions during parsing?  Some performance improvements, I suppose.

tkrodriguez:
I think you're missing that most of the method substitutions continue to work as they did before by doing graph replacement.  Look for uses of registerMethodSubstitution.  A few have been converted to directly building IR, presumably for efficiency.  You are still free to do MethodSubstitutions as they were being done before.

dougxc:
@christhalinger The compile time cost of the old way is indeed a factor (parsing bytecode is more expensive that manual node creation) but more importantly, it did not support intrinsification to be performed during bytecode parsing, only during incremental inlining.

One solution to keep the simplicity of the old way with the benefits of the new way would be to have a processor for `@MethodSubstitution` that generates the code to do the manual node creation. This would be something like the [BCELifier](https://commons.apache.org/proper/commons-bcel/xref/org/apache/bcel/util/BCELifier.html#L37). Would make for an interesting project if someone has the time ;-)

christhalinger:
@tkrodriguez Yes, I was missing that.  Thanks for pointing that out.  How do you decide which are converted to directly building IR?  Back to the `ConstantPoolSubstitutions`, does it really make sense to convert these?

christhalinger:
@dougxc Sounds like a job for @tkrodriguez 😉

tkrodriguez:
I'm not really sure you could build a BCELifier for Graal given the complexity of BytecodeParser itself plus the plugins.  Going from final IR back to code to generate it won't incorporate all sorts of flags and VM settings.

Sounds more like a job for truffle.  Don't we just want to partially evaluate BytecodeParser?  :)

dougxc:
@christhalinger I just added some [documentation](https://github.com/dougxc/graal-core/commit/5fdb347e84f62b21078ef4ef34d7a24bccf6f05f) to clarify when to use method substitution versus manual node creation that answers your question about `ConstantPoolSubstitutions`.

dougxc:
@christianwimmer please don't merge yet - there are more changes coming based on your feedback. I'll remove the `do not merge` label once they're in.

PR #17, opened 2016-02-11
improve travis gate

rschatz:
- use binary imports to reduce build time
- run unit tests in hosted mode, now that we don't need extra build time for that anymore

~~DO NOT MERGE YET~~
~~This relies on a not yet merged mx feature!~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,15 +35,8 @@ install:
   - wget https://lafo.ssw.uni-linz.ac.at/slavefiles/gate/eclipse-jdk8-linux-x86_64.tar.gz -O $ECLIPSE_TAR
   - tar -C $TRAVIS_BUILD_DIR/.. -xf $ECLIPSE_TAR
   - export ECLIPSE_EXE=$TRAVIS_BUILD_DIR/../eclipse/eclipse
+  - export MX_BINARY_SUITES=jvmci,truffle
 script:
   - mx --strict-compliance gate --strict-mode --simple
 after_failure:
-  - cat $TRAVIS_BUILD_DIR/jvmci/client-product.log
-  - cat $TRAVIS_BUILD_DIR/jvmci/server-fastdebug.log
-  - cat $TRAVIS_BUILD_DIR/jvmci/server-nojvmci-product.log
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -27,7 +27,7 @@ before_install:
   - pip install pylint==1.1.0
 install:
   - export MX_PATH=$TRAVIS_BUILD_DIR/../mx
-  - git clone https://github.com/graalvm/mx.git $MX_PATH
+  - git clone https://github.com/rschatz/mx.git $MX_PATH
```
dougxc: I assume this is just temporary?

mx.graal-core/mx_graal_8.py
```diff
@@ -320,22 +320,18 @@ def compiler_gate_runner(suites, unit_test_runs, bootstrap_tests, tasks, extraVM
...
 
 def compiler_simple_gate_runner(suites, unit_test_runs, bootstrap_tests, tasks, extraVMarguments=None):
-    # Run unit tests on server-hosted-jvmci
-    with VM('jvmci', 'fastdebug'):
-        # Build
-        with Task('BuildHotSpotJVMCI: fastdebug', tasks) as t:
-            if t: buildvms(['--vms', 'jvmci', '--builds', 'fastdebug'])
+    # Build
+    with Task('BuildHotSpotJVMCI: fastdebug', tasks) as t:
+        if t: buildvms(['--vms', 'jvmci,server', '--builds', 'product,fastdebug'])
```
dougxc: Does this actually do any building if the binary dependencies are available? Maybe we should instead check that the binaries exist and fail if not since Travis has a spotty history in terms of building certain binaries anyway.

.travis.yml
rschatz: Of course, I'll remove this once I'm finished...

mx.graal-core/mx_graal_8.py
rschatz: This does actually do something. It's building the full JDK from the system jdk and the libjvm.so from the downloaded jvmci binary dependency.

rschatz:
I've also converted the gate to use the new tags feature of mx, instead of the `--simple` flag.
Now that the relevant mx changes are in master, this is ready to be merged (assuming travis passes).

PR #21, opened 2016-02-11
made profileArguments and profileReturnType public

mbrantner:
This allows those functions to be used in other implementations of the OSRLoopNode. They are needed to avoid repeated deopts.


chumer:
This is useful/necessary for compiling OptimizedCallTarget without ever executing it.

chumer:
Waiting for travis but can get merged afterwards.

PR #23, opened 2016-02-12
CompiledExceptionHandlerTest should not rely on profiling information.

rschatz:
This unit test sometimes randomly failed on travis, now it should be reliable.

### DIFF
--- a/graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/GraalOptions.java
+++ b/graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/GraalOptions.java
@@ -122,11 +122,11 @@ public final class GraalOptions {
     @Option(help = "", type = OptionType.Debug)
     public static final OptionValue<Boolean> DeoptALot = new OptionValue<>(false);
 
-    @Option(help = "Stressed the code emitting explicit exception throwing code.", type = OptionType.Debug)
-    public static final StableOptionValue<Boolean> StressExplicitExceptionCode = new StableOptionValue<>(false);
+    @Option(help = "Stress the code emitting explicit exception throwing code.", type = OptionType.Debug)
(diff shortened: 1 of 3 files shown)


graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/GraalOptions.java
```diff
@@ -123,10 +123,10 @@
     public static final OptionValue<Boolean> DeoptALot = new OptionValue<>(false);
 
     @Option(help = "Stressed the code emitting explicit exception throwing code.", type = OptionType.Debug)
-    public static final StableOptionValue<Boolean> StressExplicitExceptionCode = new StableOptionValue<>(false);
+    public static final OptionValue<Boolean> StressExplicitExceptionCode = new OptionValue<>(false);
```
dougxc: The help message for this option and the one below should not be past tense. To be consistent with other options, it should be proscriptive: `help = "Stress the code ..."`

rschatz: Ok, I'll change that.

tkrodriguez:
The fix looks fine but I'm not sure of the point of this test if we're simply bypassing the profiling information.  It's just testing the parser now instead of checking that a profiled exception throw properly trigger that logic in the parser.  Do we need a test for the latter?

rschatz:
I interpreted these tests to check whether explicit exception edges work correctly. The assertion in `parseEager` makes sure that the parser did the correct thing, i.e. we're really testing the explicit exception case, but the main point of this test is that graal can sucessfully compile a graph that contains explicit exception edges.

tkrodriguez:
I'm fine with that

rschatz:
Fixed checkstyle error and squashed everything into a single commit.

PR #25, opened 2016-02-12
switch to container based travis

rschatz:
~~Based on #17.
It's in a separate PR so we can decide afterwards whether we want to merge this or the other.~~
Rebased on top of master.


rschatz:
Seems like container based travis is a lot slower than what we currently have (e.g. the checkstyle step takes 4:20 instead of 1:45), so I'm not going to bother to try to get it working.

Closing.

PR #30, opened 2016-02-15
[SPARC] Fix saved registers for Stub calls

zapster:
`Stub.getDestroyedRegisters()` returns the destroyed registers from the Stubs (the callees) point of view. Architectures which perform register renaming need to translate this to the callers perspective as it is used to set the temporaries of the call op.

~~In the proposed solution the translation is done by the a newly introduced method `Backend.viewAsCaller()` during `HotSpotForeignCallLinkageImpl.finalizeAddress()`~~

Alternatively we could change the behavior of `Stub.getDestroyedRegisters()` to return the translated set in the first place. In this case we could omit the `Backend` change. Personally I think this solution is counter-intuitive as I would expect that the destroyed registers set is from the Stubs perspective. On the other hand the method is only used for this purpose so we could rename it to make its behavior clearer.
(**Update:** this is now the proposed solution)

Comments/suggestions very welcome.

RFR: @rschatz @tkrodriguez

### DIFF
--- a/graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
+++ b/graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
@@ -22,8 +22,11 @@
  */
 package com.oracle.graal.compiler.target;
 
+import java.util.Set;
+
 import jdk.vm.ci.code.CodeCacheProvider;
 import jdk.vm.ci.code.CompiledCode;
+import jdk.vm.ci.code.Register;
 import jdk.vm.ci.code.RegisterConfig;
 import jdk.vm.ci.code.TargetDescription;
 import jdk.vm.ci.meta.ConstantReflectionProvider;
@@ -133,5 +136,12 @@ public abstract class Backend implements TargetProvider {
      *            installed. This argument can be null.
      */
     public abstract void emitCode(CompilationResultBuilder crb, LIR lir, ResolvedJavaMethod installedCodeOwner);
+
+    /**
+     * Translates a set of registers from the callee's perspective to the caller's perspective. This
+     * is needed for architectures where input/output registers are renamed during a call (e.g.
+     * register windows on SPARC). Registers which are not visible by the caller are removed.
+     */
+    public abstract Set<Register> translateToCallerRegisters(Set<Register> calleeRegisters);
 
 }
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotBackend.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotBackend.java
@@ -205,15 +205,15 @@ public abstract class HotSpotBackend extends Backend implements FrameMap.Referen
      * @param lir the LIR to examine
      * @return the registers that are defined by or used as temps for any instruction in {@code lir}
      */
-    protected static Set<Register> gatherDefinedRegisters(LIR lir) {
-        final Set<Register> definedRegisters = new HashSet<>();
+    protected final Set<Register> gatherDestroyedCallerRegisters(LIR lir) {
+        final Set<Register> destroyedRegisters = new HashSet<>();
         ValueConsumer defConsumer = new ValueConsumer() {
 
             @Override
             public void visitValue(Value value, OperandMode mode, EnumSet<OperandFlag> flags) {
                 if (ValueUtil.isRegister(value)) {
                     final Register reg = ValueUtil.asRegister(value);
-                    definedRegisters.add(reg);
+                    destroyedRegisters.add(reg);
                 }
             }
         };
@@ -227,7 +227,7 @@ public abstract class HotSpotBackend extends Backend implements FrameMap.Referen
                 }
             }
         }
-        return definedRegisters;
+        return translateToCallerRegisters(destroyedRegisters);
     }
 
     /**
@@ -247,7 +247,7 @@ public abstract class HotSpotBackend extends Backend implements FrameMap.Referen
      *            slot to a frame slot index
      */
     protected void updateStub(Stub stub, Set<Register> destroyedRegisters, Map<LIRFrameState, SaveRegistersOp> calleeSaveInfo, FrameMap frameMap) {
-        stub.initDestroyedRegisters(destroyedRegisters);
+        stub.initDestroyedCallerRegisters(destroyedRegisters);
 
         for (Map.Entry<LIRFrameState, SaveRegistersOp> e : calleeSaveInfo.entrySet()) {
             SaveRegistersOp save = e.getValue();
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/Stub.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/Stub.java
@@ -87,28 +87,28 @@ public abstract class Stub {
     protected CompilationResult compResult;
 
     /**
-     * The registers destroyed by this stub.
+     * The registers destroyed by this stub (from the caller's perspective).
      */
-    private Set<Register> destroyedRegisters;
+    private Set<Register> destroyedCallerRegisters;
 
-    public void initDestroyedRegisters(Set<Register> registers) {
+    public void initDestroyedCallerRegisters(Set<Register> registers) {
         assert registers != null;
-        assert destroyedRegisters == null || registers.equals(destroyedRegisters) : "cannot redefine";
-        destroyedRegisters = registers;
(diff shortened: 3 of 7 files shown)


tkrodriguez:
I think it should be returning an answer relative to the caller in all cases.  I think it should return the translated set and maybe rename it to getCallerSaveRegisters()?  I think that has the right sense.

zapster:
Thanks @tkrodriguez for the review. I like your suggestion and will update the PR.

sanzinger:
I agree, it should return the translated set, from callers perspective. I'd suggest to rename it to getDestroyedCallerRegisters - it suggest that this callee destroys the returned set of registers in the caller.

rschatz:
Code lgtm.

I don't have a particularly strong opinion on the method name, both are fine for me, but we should make it really clear in the Javadoc what these methods do and from what point of view it returns the registers.

zapster:
I'm also fine with both names. @dougxc, we need a tiebreaker. :-)

zapster:
I've decided to go with `getDestroyedCallerRegisters`. We can change the name later if we want. Since this is a bug we should try to get the code in and fix "style" issues later.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
```diff
@@ -134,4 +135,13 @@ public abstract CompilationResultBuilder newCompilationResultBuilder(LIRGenerati
      */
     public abstract void emitCode(CompilationResultBuilder crb, LIR lir, ResolvedJavaMethod installedCodeOwner);
 
+    /**
+     * Translates a register from the callees perspective to the callers perspective. This is needed
```
dougxc: callee's
caller's

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
```diff
@@ -134,4 +135,13 @@ public abstract CompilationResultBuilder newCompilationResultBuilder(LIRGenerati
...
+     * for architectures where input/output registers are renamed during a call (register windows).
```
dougxc: (e.g., register windows on SPARC)

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/Stub.java
```diff
@@ -87,28 +87,28 @@
     protected CompilationResult compResult;
 
     /**
-     * The registers destroyed by this stub.
+     * The registers destroyed by this stub (from the callers perspective).
```
dougxc: caller's

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/Stub.java
```diff
@@ -87,28 +87,28 @@
...
-        assert destroyedRegisters == null || registers.equals(destroyedRegisters) : "cannot redefine";
-        destroyedRegisters = registers;
+        assert destroyedCallerRegisters == null || registers.equals(destroyedCallerRegisters) : "cannot redefine";
+        destroyedCallerRegisters = registers;
     }
 
     /**
-     * Gets the registers defined by this stub. These are the temporaries of this stub and must thus
-     * be caller saved by a callers of this stub.
+     * Gets the registers destroyed by this stub from a callers perspective. These are the
```
dougxc: caller's

dougxc:
I prefer the name `getDestroyedCallerRegisters` as it conveys more about the action needed for such registers.

zapster:
Thanks for the review, @dougxc. PR updated.

tkrodriguez:
That name is fine with me.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
```diff
@@ -134,4 +135,14 @@ public abstract CompilationResultBuilder newCompilationResultBuilder(LIRGenerati
...
 
+    /**
+     * Translates a register from the callee's perspective to the caller's perspective. This is
+     * needed for architectures where input/output registers are renamed during a call (e.g.
+     * register windows on SPARC).
+     *
+     * @return the translated {@link Register} or {@code null} if the {@code calleeRegister} is not
+     *         visible from the caller.
+     */
+    public abstract Register asCallerSavedRegister(Register calleeRegister);
```
tkrodriguez: Why is the word "saved" in this name?  It doesn't have any relationship to saving really.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotBackend.java
```diff
@@ -205,15 +205,17 @@ public void completeInitialization(HotSpotJVMCIRuntime jvmciRuntime) {
...
+    protected final Set<Register> gatherDestroyedCallerRegisters(LIR lir) {
+        final Set<Register> destroyedRegisters = new HashSet<>();
         ValueConsumer defConsumer = new ValueConsumer() {
 
             @Override
             public void visitValue(Value value, OperandMode mode, EnumSet<OperandFlag> flags) {
                 if (ValueUtil.isRegister(value)) {
-                    final Register reg = ValueUtil.asRegister(value);
-                    definedRegisters.add(reg);
+                    final Register reg = asCallerSavedRegister(ValueUtil.asRegister(value));
```
tkrodriguez: I don't understand the purpose of this code.  Registers seen in a LIR op should always be in terms of the caller so they should never require translation.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
zapster: what do you suggest? just `getCallerRegister` or `toCallerRegister`?

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotBackend.java
zapster: Well, this code sees the registers as they are in the `LIR` aka as seen by the Stub. But callers of the Stub need a translation.

tkrodriguez: Ok.  I see where we are. I was having trouble making sense of where this code was.  We're building the stub and collecting all registers which are defined.  Then to produce the answer to getDestroyedCallerRegisters we need to filter out all callee saved registers and convert the registers to the callers notion of the names.  asCallerSavedRegister is doing both of those things.  I think it would be better to have a single method that takes the set of destroyed registers and does that conversion on the set.  The default one does nothing and sparc has an override that does the filtering.  asCallerSavedRegister is just a weird helper to have lying around to me.

zapster: Ok, this is what the original solution was about. The downside if this is that we are creating the destroyed set, filter it and create the destroyed-caller set only to transform it into the Value array.

tkrodriguez: It can certainly just return a Value[] instead.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
```diff
@@ -134,4 +137,11 @@ public abstract CompilationResultBuilder newCompilationResultBuilder(LIRGenerati
      */
     public abstract void emitCode(CompilationResultBuilder crb, LIR lir, ResolvedJavaMethod installedCodeOwner);
 
+    /**
+     * Translates a set of registers from the callee's perspective to the caller's perspective. This
+     * is needed for architectures where input/output registers are renamed during a call (e.g.
+     * register windows on SPARC).
+     */
+    public abstract Set<Register> getCallerSavedRegisters(Set<Register> calleeRegisters);
```
zapster: @tkrodriguez should I rename this method? not sure if your original comment about the name is still valid.

tkrodriguez: I would leave out the Saved part.  Or maybe translateToCallerRegisters?  Also note in the javadoc that registers may also be removed because of things like register windows.

tkrodriguez:
This version looks right to me.  Sorry if I confused you with my earlier comments.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/target/Backend.java
zapster: Ok, will do.

zapster:
@tkrodriguez no worries. Thanks for reviewing!

PR #31, opened 2016-02-15
Add jdk9 to test matrix.

rschatz:
This runs the gate on our jdk9 snapshot in parallel to the normal gate. The jdk9 gate skips the style checks (checkstyle, findbugs, ...), they would be redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,23 +20,40 @@ dist: trusty
 env:
   global:
     - DEFAULT_VM="jvmci"
-before_install:
-  - virtualenv venv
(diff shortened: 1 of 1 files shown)


dougxc:
Looks promising - now you just need to get the gate passing on jdk9 ;-)

rschatz:
Interesting... In my fork it passed. I'll look into it.

rschatz:
I can't reproduce the error locally... The error itself looks like a confusion between the compiler and the test whether G1 or the SerialGC is in use.

PR #32, opened 2016-02-15
respect -XX:-Inline

tkrodriguez:
Is there a better place to plug into this?  Some hotspot tests only work if this flag is respected.

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompilationTask.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompilationTask.java
@@ -47,6 +47,7 @@ import jdk.vm.ci.runtime.JVMCICompiler;
 import jdk.vm.ci.services.Services;
 
 import com.oracle.graal.code.CompilationResult;
+import static com.oracle.graal.compiler.phases.HighTier.Options.Inline;
 import com.oracle.graal.debug.Debug;
 import com.oracle.graal.debug.Debug.Scope;
 import com.oracle.graal.debug.DebugCloseable;
@@ -55,6 +56,8 @@ import com.oracle.graal.debug.DebugMetric;
 import com.oracle.graal.debug.DebugTimer;
 import com.oracle.graal.debug.Management;
 import com.oracle.graal.debug.TTY;
+import com.oracle.graal.options.OptionValue;
+import com.oracle.graal.options.OptionValue.OverrideScope;
 
 //JaCoCo Exclude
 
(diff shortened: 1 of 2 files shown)


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompilationTask.java
```diff
@@ -178,7 +181,11 @@ public CompilationRequestResult runCompilation() {
             try (Scope s = Debug.scope("Compiling", new DebugDumpScope(String.valueOf(getId()), true))) {
                 // Begin the compilation event.
                 compilationEvent.begin();
-                result = compiler.compile(method, entryBCI, useProfilingInfo);
+                // Use HotSpot's Inline flag if Graal's flag has the default setting
+                boolean inline = Inline.hasDefaultValue() ? config.inline : Inline.getValue();
```
dougxc: Should we issue some kind of warning if `-XX:+Inline -G:-Inline` or any other combination of incompatible settings? Or do we just say an explicit Graal setting overrides any HotSpot setting?

dougxc:
I can't think of a better place since where you've put means it will work for CompileTheWorld compilations as well.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompilationTask.java
tkrodriguez: I'm not sure what's reasonable here either.  Warning might be a good idea.  But we're also only overriding it for compiles triggered from HotSpot.  Truffle for instance ignore -XX:-Inline and I'm not sure that's right.

dougxc: At some point, we need to re-design the option system to make certain options less global without relying on the hacky `OverrideScope` mechanism. That said, I think Truffle ignoring `-XX:-Inline` is right since there should really be a `-G:-TruffleGraalInline` option for that use case.

tkrodriguez:
Should I add a warning or is this fine as is?

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompilationTask.java
woess: There are actually three kinds of inlining that can happen in Truffle currently:
1. the forced deterministic inlining done by the partial evaluation phase (unaffected by flags).
2. Truffle inlining: inlining decisions taken entirely by Truffle, then performed by 1. (-G:-TruffleFunctionInlining)
3. optional Graal inlining that still happens after the partial evaluation phase (if any). (-G:-Inline, plus after this change -XX:-Inline)

Since it doesn't make much sense to do PE without inlining, I believe this is all right.

tkrodriguez: Agreed.

PR #36, opened 2016-02-16
updated CONTRIBUTING.md

dougxc:


CONTRIBUTING.md
```diff
@@ -1,2 +1,16 @@
-Please see http://openjdk.java.net/contribute/ for details of how to contribute
-to OpenJDK projects including GraalVM.
+To submit pull requests to graal-core, you need to sign the [Oracle Contributor
+Agreement][1].
+
+All pull requests should be given an [Assignee][2]. The assignee is
+responsible for adding the `accept` label to the pull request once they have
+completed a review and are satisfied that all review comments have been addressed.
+After the `accept` label has been applied, the pull request author can merge the
+pull request if they have the necessary permissions. Otherwise, the assignee will
```
zapster: In order to speed things up I think everyone with the right permissions should be allowed to merge if there is an `accept` label. The assignee should feel responsible, though.

rschatz: No, actually I don't think that's a good idea. The creator might e.g. have a history cleanup planned, and deferred it until after the accept to save travis runs.

One advantage of the `accept` label policy is that the creator of the PR doesn't need to feel responsible to keep it up-to-date and clean at all times, a single merge/rebase/cleanup just before merging is sufficient.

zapster: Ok.

dougxc: Yes, in the case where the PR cannot do the merge, then someone else with the right permissions should. I'll update to reflect that.

dougxc: @rschatz in that case, the PR creator should add a `do not merge` label

dougxc: ... which of course they can't if they don't have permission to adjust labels in the first place.

dougxc: For pull requests submitted by external contributors (those without wrote permission to the `graal-core` repo), the onus should be on the contributor to keep the PR up-to-date and clean and on the reviewer(s) to encourage the contributor to do so (e.g., by not accepting the PR until such cleanups have been performed). This removes the need for the merger to synchronize with the contributor once the `accept` label has been applied.

woess: +1

rschatz: Ok, we can do this with the `do not merge` label.

I can see the value of others being allowed to merge my pull requests... (since one of my travis builds just got spontaneously restarted and now I have to leave office before being able to press the merge button :P)

PR #39, opened 2016-02-17
Graal runtime adoptions for the new instrumentation framework.

chumer:


graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
```diff
@@ -629,4 +641,14 @@ public final int hashCode() {
     void setCompilationTask(Future<?> compilationTask) {
         this.compilationTask = compilationTask;
     }
+
+    static final AccessorOptimizedCallTarget ACCESSOR = new AccessorOptimizedCallTarget();
```
woess: why not private?

PR #42, opened 2016-02-17
[AArch64] AArch64FloatArithmeticSnippets: Fix remaining failures and …

christhalinger:
…add JVMS language to explain the tests.

### DIFF
--- a/graal/com.oracle.graal.replacements.aarch64/src/com/oracle/graal/replacements/aarch64/AArch64FloatArithmeticSnippets.java
+++ b/graal/com.oracle.graal.replacements.aarch64/src/com/oracle/graal/replacements/aarch64/AArch64FloatArithmeticSnippets.java
@@ -73,38 +73,52 @@ public class AArch64FloatArithmeticSnippets extends SnippetTemplate.AbstractTemp
 
     @Snippet
     public static float fremSnippet(float x, float y) {
+        // JVMS: If either value1' or value2' is NaN, the result is NaN.
+        // JVMS: If the dividend is an infinity or the divisor is a zero or both, the result is NaN.
         if (Float.isInfinite(x) || y == 0.0f || Float.isNaN(y)) {
             return Float.NaN;
         }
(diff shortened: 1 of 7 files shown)


graal/com.oracle.graal.replacements.aarch64/src/com/oracle/graal/replacements/aarch64/AArch64FloatArithmeticSnippets.java
```diff
@@ -73,38 +73,52 @@ public void lower(RemNode node, LoweringTool tool) {
...
         }
-        return safeRem(x, y);
+        return result;
     }
 
     @Snippet
     public static double dremSnippet(double x, double y) {
+        // JVMS: If either value1' or value2' is NaN, the result is NaN.
+        // JVMS: If the dividend is an infinity or the divisor is a zero or both, the result is NaN.
         if (Double.isInfinite(x) || y == 0.0 || Double.isNaN(y)) {
```
tkrodriguez: Shouldn't there be a check that x is not NaN?

christhalinger: Not sure.  It might be that the machine instruction gets that one right.  This code is from Daniel.

Maybe we should have a check per JVMS rule.  Would be easier to understand and the compiler should fold the tests anyway.

tkrodriguez: I'd expect that either both are needed or neither is.  Presumably this is being exercised by the new test cases though.

christhalinger: It is and they pass so I think it's fine :)

PR #43, opened 2016-02-17
CTW exclude filters should only control compilation

tkrodriguez:
The method filters can help reproducing a failure quickly but they also change which classes are loaded which impairs reproducibility.  Instead we should perform exactly the same loading but stop before compiling them.


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/CompileTheWorld.java
```diff
@@ -583,6 +576,14 @@ public GraalDebugConfig getDebugConfig() {
                             println("Preloading failed for (%d) %s: %s", classFileCounter, className, t);
                         }
 
+                        // Load and resolve classes even if we aren't compiling them
```
dougxc: Better comment: Only check filters after class loading and resolution to mitigate impact on reproducibility.

PR #44, opened 2016-02-18
[AArch64] Make compare-with-immediate work correctly. This also makes…

christhalinger:
… other inline constants work, like for loads.


christhalinger:
This branch is screwed up.  Opened #49 instead.

PR #45, opened 2016-02-18
parallel gate

rschatz:
Parallelizes the gate into 6 separate runs:
- style checks
- ecj build / warnings / findbugs
- unit tests on jdk8
- bootstrap on jdk8
- unit tests on jdk9
- bootstrap on jdk9

Only the `BuildJavaWithJavac` step is running in 5 of these 6 gates, everything else is non-overlapping.

~~DO NOT MERGE YET, this is relying on a not yet merged mx feature.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,8 +55,8 @@ install:
   - |
       if [ "$JDK" == "jdk9" ]
       then
-        JDK_TAR=$TRAVIS_BUILD_DIR/../jdk9-SNAPSHOT-fastdebug-linux-amd64.tar.gz
-        wget https://lafo.ssw.uni-linz.ac.at/slavefiles/jdk/jdk9-SNAPSHOT-fastdebug-linux-amd64.tar.gz -O $JDK_TAR
+        JDK_TAR=$TRAVIS_BUILD_DIR/../jdk9-SNAPSHOT-release-linux-amd64.tar.gz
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -22,12 +22,16 @@ env:
...
+    - GATE="fullbuild"
+    - JDK="jdk8" GATE="build,test"
+    - JDK="jdk8" GATE="build,bootstrap"
     - JDK="jdk9" GATE="build,test"
+    - JDK="jdk9" GATE="build,bootstrap"
 install:
   - |
       export MX_PATH=$TRAVIS_BUILD_DIR/../mx
-      git clone https://github.com/graalvm/mx.git $MX_PATH
+      git clone https://github.com/rschatz/mx.git -b fullbuild-tag $MX_PATH
```
rschatz: This change will be removed once the corresponding mx change is merged.

rschatz:
Timeline for the last travis run:
- after the push, the first 5 tasks started simultaneously
- the first task (style) finished after 5 minutes
- the 6th task started about 30 seconds later
- total run time about 15 minutes

Based on that, I think we should merge the `style` and `fullbuild` tasks. Both together should be still faster than the currently slowest task (jdk8 unittests), and then we're down to 5 builders in parallel. That should shave off another minute or two from total gate time.

dougxc:
This is a nice improvement (assuming Travis reliably provides this level of parallelism).

rschatz:
The latest travis run went roughly as expected:
- first 4 tasks started simultaneously
- 5th task started about 1 minute late (but didn't finish last anyway)
- total run time about 14 minutes

PR #48, opened 2016-02-18
[AArch64] Fix AArch64Move.stack2reg for floats.

christhalinger:


tkrodriguez:
Please be careful about stacking PRs on top of each other.  It makes it hard to review since it's unclear which changes are actually new.

christhalinger:
I think this happened because I started the branch from another branch and not master.

PR #52, opened 2016-02-19
Negate node must not canonicalize input subtraction on FP stamps

sanzinger:
Canonicalizing negates of subtractions is not allowed in FP

-1.0-(-1.0)==0.0 but -(-1.0-(-1.0))==-0.0


rschatz:
Stupid -0.0, but I don't see better a way to optimize this either...

PR #55, opened 2016-02-19
Use ArithmeticOperation as common superinterface for nodes that use the ArithmeticOpTable mechanism.

rschatz:
There was no common superinterface for the nodes using the `ArithmeticOpTable` mechanism of type-independent stamp and constant folding, and therefore no way to get the `Op` descriptor.

The old `ArithmeticOperation` interface is a nice place to put this. This means that some nodes that used to be `ArithmeticOperation` are now not (e.g. `AMD64MathIntrinsicNode`). On the other hand, the old `ArithmeticOperation` interface was redundant anyway, there was not a single class that implemented `ArithmeticOperation` but not its subinterface `ArithmeticLIRLowerable`.

The new design is more flexible, compiler phases can now directly query the `Op` from nodes that support it, instead of relying on the functionality explicitly exposed by the node via special methods like `foldStamp` or `tryConstantFold`, in case we want to add more type-dependent optimizations.

/cc @dougxc

### DIFF
--- a/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/ArithmeticOperation.java
+++ b/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/ArithmeticOperation.java
@@ -22,9 +22,13 @@
  */
 package com.oracle.graal.nodes;
 
+import com.oracle.graal.compiler.common.type.ArithmeticOpTable.Op;
+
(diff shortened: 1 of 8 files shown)


dougxc:
lgtm

lukasstadler:
lgtm, please merge

PR #56, opened 2016-02-22
Make CompiledExceptionHandlerTest more robust.

rschatz:
Sometimes the test still failed because of random inlining decisions creating a second execption handler (e.g. in one of the StringBuilder methods).


rschatz:
Somehow I can still make the unit test fail under some circumstances. It even makes a different whether the test is compiled with `javac` or `ecj`. I have to investigate further, in the meanwhile I'm putting this PR on hold until I fully understand the issue.

rschatz:
Ok, the remaining bug was that sometimes there was a stale value in the `lastInlineInfo` field in the `BytecodeParser`. I have refactored the whole thing to get rid of this field, threading the value through arguments instead. Perfect example of why you shouldn't use global variables :)

PR #57, opened 2016-02-22
Overriding of -XX:-Inline with Graal options isn't working properly

tkrodriguez:
This seems to be the cause of the ctw failures.  The changes to respect -XX:-Inline relied on hasDefaultValue but CTW overrides the Graal inline flag when run from the gate.  We weren't detecting that override anymore so compiling everything with inlining.  We still have debug why it fails.

### DIFF
--- a/graal/com.oracle.graal.options/src/com/oracle/graal/options/OptionValue.java
+++ b/graal/com.oracle.graal.options/src/com/oracle/graal/options/OptionValue.java
@@ -86,9 +86,7 @@ public class OptionValue<T> {
             Entry<OptionValue<?>, Object> single = overrides.entrySet().iterator().next();
             OptionValue<?> option = single.getKey();
             Object overrideValue = single.getValue();
-            if (!overrideValue.equals(option.getValue())) {
-                return new SingleOverrideScope(option, overrideValue);
-            }
+            return new SingleOverrideScope(option, overrideValue);
         }
         return new MultipleOverridesScope(current, overrides);
     }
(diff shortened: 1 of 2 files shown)


tkrodriguez:
I realized that hasDefaultValue doesn't really work the way I expect.  I assumed it meant "has this flag been explicitly set in any manner" but the the way overrides work it drops the override if the value happens to be the same which defeats that meaning.  I've changed the meaning but I'm curious whether you think that's the right behaviour.  Or maybe I should add a new test like hasBeenSet instead.

graal/com.oracle.graal.options/src/com/oracle/graal/options/OptionValue.java
```diff
@@ -249,13 +246,22 @@ public T getDefaultValue() {
     }
 
     /**
-     * Returns true if the option has the same value that was set in the source code.
+     * Returns true if the option has not been set from the command line. Note that this doesn't
+     * mean that the current value is different than the default.
      */
```
dougxc: Sounds like we need another name for this method. I would adopt the `hasBeenSet` suggestion you made.

@christianwimmer @olpaw , does that impact your usages of this API?

tkrodriguez: OK I'll add the new method.  But it seem like hasDefaultValue needs to respect overrides.  Why isn't it simply Object.equals(getValue(), getDefaultValue())?

tkrodriguez:
Ok, I've added hasBeenSet and fixed getDefaultValue respect overrides.

PR #61, opened 2016-02-23
Add support for reserved stack access checks from JDK9

tkrodriguez:
This is some ugly new code that's part of JDK9.  http://openjdk.java.net/jeps/270 describes the idea and https://bugs.openjdk.java.net/browse/JDK-8046936 is the bug it was integrated under.  Basically we need a little snippet of code that runs just before returning from a method that's inlined a method that's been annotation with ReservedStackAccess.  There's apparently no requirement that we execute this logic if an exception is being propagated through the code.  JDK8 has a stub implementation of HotSpotResolvedJavaMethod.hasReservedStackAccess to simplify things a little bit.


tkrodriguez:
Maybe @dougxc and @gilles-duboscq could take a look too

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/phases/ReservedStackAccessCheckPhase.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    protected void run(StructuredGraph graph) {
+        if (graph.method() != null) {
+            boolean reserved = hasReservedStackAccess(graph.method());
+            for (ResolvedJavaMethod method : graph.getInlinedMethods()) {
+                reserved = reserved || hasReservedStackAccess(method);
+            }
+            if (reserved || true) {
```
gilles-duboscq: `|| true`? some left-over debug?

tkrodriguez: Oops.  Yes I was forcing it to run in JDK8 at one point.  The guards I added to protect adding the phase to low tier would have kept it from reaching this code now.

rschatz:
Is there a some way to test this reliably from a unit test? What happens if we get that stuff wrong?

dougxc:
Can you please update CHANGELOG.md to mention this change.

tkrodriguez:
I'll update the changelog.

We can't really test this in JDK8 since the support doesn't exist.  In JDK9 there's a regression tests which exercises this and I've tested it there at least on AMD64.  I still need to check the code on sparc.

One thing that concerns me when I look at the code we generate for the runtime calls is that I don't see any explicit logic for preserving r15 around the native call.  It doesn't seem to get destroyed but I'm not sure how that's supposed to work.

dougxc:
r15 is [callee saved](https://github.com/graalvm/graal-core/blob/36d28f085bd3548a9e19d6d9ae31de5f28605e2c/graal/com.oracle.graal.hotspot.amd64/src/com/oracle/graal/hotspot/amd64/AMD64HotSpotBackendFactory.java#L213-L218) so there should not be an issue.

rschatz:
Then we should at least add a unit test that runs only on JDK9.

tkrodriguez:
The annotation is in jdk.internal.vm.annotation so it's not going to be accessible once modules are in place, at least without using special module access flags.  Can we have unit tests that are only built if jdk9 is available?

ReentrantLock and AbstractQueuedSynchronizer both use the annotation.  I could write a generic test that exercises one of those and tries to induce StackOverflows to exercise it.  Would that do?

tkrodriguez:
Ok r15 is ok then but there might be a lurking WIN64 issue.

`void MacroAssembler::call_VM_leaf_base(address entry_point, int num_args) {
  Label L, E;

#ifdef _WIN64
  // Windows always allocates space for it's register args
  assert(num_args <= 4, "only register arguments supported");
  subq(rsp,  frame::arg_reg_save_area_bytes);
#endif

  // Align stack if necessary
  testl(rsp, 15);
  jcc(Assembler::zero, L);

  subq(rsp, 8);
  {
    call(RuntimeAddress(entry_point));
  }
  addq(rsp, 8);
  jmp(E);

  bind(L);
  {
    call(RuntimeAddress(entry_point));
  }

  bind(E);

#ifdef _WIN64
  // restore stack pointer
  addq(rsp, frame::arg_reg_save_area_bytes);
#endif

}`

I think we always run with 16 byte alignment in generated code so we should be ok there.  Do we have the WIN64 logic anywhere?

gilles-duboscq:
@tkrodriguez regarding 9-specific unit tests, you can use [`Assume`](http://junit.sourceforge.net/javadoc/org/junit/Assume.html) to have check preconditions in your test.

tkrodriguez:
I think it would require build system support to directly use the annotation.  The file itself wouldn't compile in JDK8.

I'd actually prefer no to add a test for this to our source base.  The ReservedStackAccess implementation in hotspot is fairly crap in my estimation and I'd prefer to just rely on their test case to find any issues.  It's on my todo list to file a bug expressing my thoughts on what's wrong.  The whole point of the annotation for unmodified code to take advantage of it but the test case only works with inlining turned off.  C2 fails the test if inlining is enabled as do we after my fixes.

PR #64, opened 2016-02-24
Disable BC_{df}rem unittests on SPARC temporarily

sanzinger:
Issue #30 will be fixed soon, until then, I've disabled the offending unittests.


dougxc:
With this change, does the gate pass on SPARC?

sanzinger:
Unfortunately not, there is another issue #65

dougxc:
That should have been addressed by #66. Please try the SPARC gate again after merging #66 into this branch. That said, it's fine to merge this PR even if new failures turn up. Just file new issues for the latter.

PR #67, opened 2016-02-24
Add CTW to travis gate

tkrodriguez:
This adds CTW into the travis gate.


tkrodriguez:
I want to see what this does to build times.

tkrodriguez:
If the gate times are in a reasonable range I'd like to push it but I suspect they won't be.

tkrodriguez:
So adding ctw to 8 adds 11 minutes and for 9 it adds 13 minutes.  These become the longest part of the build at 22 minutes as opposed to the old build which was 15 minutes I think.

PR #68, opened 2016-02-24
Try containers again

tkrodriguez:


tkrodriguez:
I was curious whether there was a difference in promptness of starting container based build since according to the status there are usually lots more of those running while the regular linux builds seem to saturate during a lot of the day.  Didn't seem to make any difference.  The build times are definitely worse.  A regular build takes 12,9,14,8,10 minutes but the container build took 17,13,15,9,11 minutes and findbugs failed.  Something about the first task takes quite a bit longer in this config.  The others aren't that much worse.

PR #70, opened 2016-02-24
Allow getting the host's StampProvider through GraalRuntime.getCapability

gilles-duboscq:


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotGraalRuntime.java
```diff
@@ -201,6 +203,8 @@ public String getName() {
             return (T) this;
         } else if (clazz == SnippetReflectionProvider.class) {
             return (T) getHostProviders().getSnippetReflection();
+        } else if (clazz == StampProvider.class) {
+            return (T) new HotSpotStampProvider();
```
rschatz: Shouldn't this be `(T) getHostProviders().getStampProvider()`?
This would allow e.g. a platform to override the stamp provider (currently not needed, but you never know).

gilles-duboscq: Good point, i'll do that

dougxc:
lgtm

PR #71, opened 2016-02-24
Add separate CTW tasks

tkrodriguez:
Instead of adding CTW to the other tasks create separate tasks for it.


tkrodriguez:
This seems worse than adding them to an existing task

PR #73, opened 2016-02-25
Increase parallelism in gate

tkrodriguez:
force max-cpus to 3 to see how it affects the gate


tkrodriguez:
No visible effect on the build time

PR #74, opened 2016-02-25
Try caching the mx directory during travis builds

tkrodriguez:
See if the caching effects can be seen through a pull request.


tkrodriguez:
It looks to me like this does nothing until the master repo has caching enabled so we'd need to push this if we wanted to see the effects of it.  Unless I've somehow incorrectly setup the caching.

tkrodriguez:
Hmm. This is only available for private repos and container based builds.

tkrodriguez:
Doesn't seem helpful

PR #75, opened 2016-02-25
CompileTheWorld now includes class initializers (i.e., <clinit>).

dougxc:
While most uses of Graal won't ever compile class initializers, it's still possible and so we should stress it.


tkrodriguez:
looks good

PR #76, opened 2016-02-25
Fix/dominator ce

davleopo:
(1) Remove some calls to debug.log & cleanup of an old and invalid call to debug(string.format(...))
(2) Remove toString() calls of parameters of Debug.log("%s",obj) calls to avoid having them evaluated even if debugging is disabled

### DIFF
--- a/graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/DominatorConditionalEliminationPhase.java
+++ b/graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/DominatorConditionalEliminationPhase.java
@@ -849,7 +849,6 @@ public class DominatorConditionalEliminationPhase extends Phase {
         private final ValueNode guard;
 
         public InfoElement(Stamp stamp, ValueNode guard) {
-            Debug.log(4, "Creating an info element with stamp %s and guard %s", stamp.toString(), guard.toString());
             this.stamp = stamp;
             this.guard = guard;
         }
--- a/graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/DominatorConditionalEliminationPhase.java
+++ b/graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/DominatorConditionalEliminationPhase.java
@@ -102,7 +102,6 @@ public class DominatorConditionalEliminationPhase extends Phase {
     @Override
     @SuppressWarnings("try")
     protected void run(StructuredGraph graph) {
-        Debug.log("Starting conditional elimination....");
         try (Debug.Scope s = Debug.scope("DominatorConditionalElimination")) {
             Function<Block, Iterable<? extends Node>> blockToNodes;
             Function<Node, Block> nodeToBlock;
@@ -173,7 +172,7 @@ public class DominatorConditionalEliminationPhase extends Phase {
 
             @Override
             public void postprocess() {
-                Debug.log("[Post Processing block %s]", block.toString());
+                Debug.log("[Post Processing block %s]", block);
                 undoOperations.forEach(x -> x.run());
             }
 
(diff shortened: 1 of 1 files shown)


graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/DominatorConditionalEliminationPhase.java
```diff
@@ -173,7 +172,7 @@ public InstanceFrame(Block block, InstanceFrame parent) {
 
             @Override
             public void postprocess() {
-                Debug.log("[Post Processing block %s]", block.toString());
```
dougxc: This kind of this should have been caught by [VerifyDebugUsage](https://github.com/graalvm/graal-core/blob/6daee4240193a87f43f6b8586f2749e4bae9816e/graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java#L38-L41). Can you please find out why it wasn't.

davleopo: yes, I will, thanks for the hint

davleopo: [VerifyDebugUsage](https://github.com/graalvm/graal-core/blob/6daee4240193a87f43f6b8586f2749e4bae9816e/graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java#L38-L41) only checks for StringBuilder and StringBuffer  [receivers](https://github.com/graalvm/graal-core/blob/6daee4240193a87f43f6b8586f2749e4bae9816e/graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java#L61) of `toString` calls in the log call 

so calls like 
`Debug.log("%s",obj.toString())` are not considered to be invalid. I guess the rationale behind that is that a `toString()` call on an arbitrary receiver cannot be counted as invalid if the `toString` method performs more complex work or has parameters?

an example would be [MatchStatement](https://github.com/graalvm/graal-core/blob/6daee4240193a87f43f6b8586f2749e4bae9816e/graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/match/MatchStatement.java#L112) where there happens a `toString` call that is on purpose and that would be considered an error else

I think it we could extend `verifyDebugUsage` to filter `toString `methods without parameters like? (but I am not sure if this invariant might be too restrictive...)

```
if (m.getSignature().getParameterCount(false) == 0 &&    m.getSignature().getReturnType(m.getDeclaringClass()).equals(
context.getMetaAccess().lookupJavaType(String.class))) {

   StackTraceElement e = graph.method().asStackTraceElement(invoke.bci());

   throw new VerificationError(String.format("%s: parameter %d of call to %s appears to be a                 
                toString call without a parameter %n" +
                "    toString calls are handled by Debug.log, do not call them on parameters", e, 
                argIdx, callee.format("%H.%n(%p)")));
}
```

dougxc: `toString()` calls with parameters have to be allowed as there's no way for `Debug.log` to make such calls itself. These usages of `Debug.log` should have a guard testing whether logging is enabled. However, all parameterless calls to `toString` should be caught by this check. If you're expecting side effects of `toString()` in a call to `Debug.log`, you're doing something very wrong.

davleopo: ok, should I prepare a patch adding the check for parameterless `toString` methods?

dougxc: yes please!

davleopo: ok, thanks for your input, will do

zapster:
Have you verified that your changes actually fix the compile time regression?

zapster:
I think we should do the VerifyDebugUsage fixes in a separate PR. @dougxc  what to you think?

davleopo:
sorry, I was unclear, I would have prepared another PR....

dougxc:
yep, separate PR - thanks!

davleopo:
@zapster I have done the testing now and this PR should fix the compile time regressions in the dominator conditional elimination phase

I will update the PR and merge

zapster:
Thanks @davleopo!

PR #77, opened 2016-02-25
fix NFI when used from the truffle partial evaluator

rschatz:
The partial evaluator didn't run the normal Graal invocation plugins like a standard Java compilation. Since the NFI plugin can't do anything useful when the function is not yet constant, it needs to be re-run during partial evaluation.

This PR also contains a unit test demonstrating the problem. It also includes an update of the truffle version, since the unit test triggers the NPE of graalvm/truffle#72.

### DIFF
--- a/graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
+++ b/graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
@@ -1534,9 +1534,10 @@ public class BytecodeParser implements GraphBuilderContext {
         return null;
     }
 
-    public void intrinsify(ResolvedJavaMethod targetMethod, ResolvedJavaMethod substitute, ValueNode[] args) {
+    public boolean intrinsify(ResolvedJavaMethod targetMethod, ResolvedJavaMethod substitute, ValueNode[] args) {
         boolean res = inline(targetMethod, substitute, true, args);
         assert res : "failed to inline " + substitute;
+        return res;
     }
 
     private boolean inline(ResolvedJavaMethod targetMethod, ResolvedJavaMethod inlinedMethod, boolean isIntrinsic, ValueNode[] args) {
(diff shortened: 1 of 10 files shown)


rschatz:
@mrigger, this should the native call performance problems in sulong.

graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java
```diff
@@ -100,6 +99,7 @@ protected boolean isObjectOfLanguage(Object object) {
         return false;
     }
 
+    @SuppressWarnings("deprecation")
     @Override
     protected Visualizer getVisualizer() {
         return null;
```
woess: we can remove this method when https://github.com/graalvm/truffle/pull/68 is merged.

woess:
@rschatz Thanks for the fix and for adding a test. LGTM.

graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java
woess: anyway, the instrumentation tests will be updated soon.

rschatz:
It's failing one of the PE tests. Apparently some of the default graph builder plugins can't run during PE because the GraphBuilderContext implementation used during PE is not complete.

I'll look into how to fix that.

graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java
rschatz: Ok. Looks like that's going to be integrated soon, so I'll wait for that change before integrating this...

rschatz:
Fixed, and also updated include graalvm/truffle#68. Please review again.
/cc @thomaswue (because of the PE changes)

rschatz:
The truffle update breaks some truffle unit tests. Due to a bug in mx this doesn't show up in the travis gate. I'm putting this on hold until both the mx and the truffle bug are fixed.

rschatz:
Ok, next try ;)

This is basically the same content as before, but with the truffle update removed, since that was merged separately in the meantime.

PR #78, opened 2016-02-25
Quick-fix README.md

### LINKED ISSUE #1: README.md is outdated
The current [README.md](README.md) still refers to [mxtool/mx.py](mxtool/mx.py) and mentions "building C++". This should be synced with the instructions on https://wiki.openjdk.java.net/display/Graal/Instructions. In order to avoid redundancy and synchronization afford I think we should keep these instructions only at one location and link to it if needed.

zapster:
No documentation is better than wrong documentation. Fixes #1.


dougxc:
Thanks for the reminder. I will now starting migrating content from https://wiki.openjdk.java.net/display/Graal to .md files in this repo. As such, I'll close this PR since README is about to become up to date.

PR #81, opened 2016-02-25
Fix problems in graal stack trace handling to match those of the default runtime in the latest version.

chumer:
This changeset should bring the graal runtime in a compatible state with the default runtime:

Here is the matching Truffle change (needs to get merged before we can push this).
https://github.com/graalvm/truffle/pull/79/files

### DIFF
--- a/graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java
+++ b/graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java
@@ -33,7 +33,6 @@ import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.instrument.ASTProber;
 import com.oracle.truffle.api.instrument.EventHandlerNode;
 import com.oracle.truffle.api.instrument.Instrumenter;
-import com.oracle.truffle.api.instrument.KillException;
 import com.oracle.truffle.api.instrument.Probe;
 import com.oracle.truffle.api.instrument.SyntaxTag;
 import com.oracle.truffle.api.instrument.Visualizer;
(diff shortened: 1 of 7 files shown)


lukasstadler:
this fails travis because of a formatting issue:
```diff
 - graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
Changes:
--- 
+++ 
@@ -328,7 +328,7 @@
     public FrameInstance getCallerFrame() {
         return iterateImpl(frame -> {
             return frame;
-        } , 1);
+        }, 1);
     }
 
     @TruffleBoundary```

**Comment by lukasstadler** [comment]
looks good to me

File: graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/InstrumentationPETestLanguage.java [hunk 1]
```diff
@@ -100,6 +99,7 @@ protected boolean isObjectOfLanguage(Object object) {
         return false;
     }
 
+    @SuppressWarnings("deprecation")
```
rschatz: Can't this simply be removed, now that it's not abstract anymore?

dougxc:
Since the graal-core gate doesn't currently run the truffle tests (since truffle is a binary dependency), can you please confirm that `mx --vm server unittest` passes after this change when truffle is a source suite dependency.

chumer:
Yes it should run. I tested it.

chumer:
@dougxc But it should definitely run the Truffle unit tests in the future. Otherwise the graal truffle runtime might miss some tests.

dougxc:
@chumer I agree: https://github.com/graalvm/mx/issues/58

Can you please take care of creating the `TRUFFLE_TEST` distribution.

PR #83, opened 2016-02-26
mx: set default graal.options.file property to graal.options.

zapster:
/cc @dougxc

### DIFF
--- a/mx.graal-core/mx_graal_8.py
+++ b/mx.graal-core/mx_graal_8.py
@@ -400,20 +400,23 @@ def jdkartifactstats(args):
         print '{:>10}  {}'.format('<missing>', jvmLib)
 
 # Support for -G: options
-def _translateGOption(arg):
-    if arg.startswith('-G:+'):
-        if '=' in arg:
-            mx.abort('Mixing + and = in -G: option specification: ' + arg)
-        arg = '-Dgraal.' + arg[len('-G:+'):] + '=true'
-    elif arg.startswith('-G:-'):
-        if '=' in arg:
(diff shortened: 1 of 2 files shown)


zapster:
Is this change worth a ChangeLog entry?

mx.graal-core/mx_graal_8.py
```diff
@@ -400,20 +400,23 @@ def _getDeployedJars():
...
+            if '=' in arg:
+                mx.abort('Mixing - and = in -G: option specification: ' + arg)
+            arg = '-Dgraal.' + arg[len('-G:+'):] + '=false'
+        elif arg.startswith('-G:'):
+            if '=' not in arg:
+                mx.abort('Missing "=" in non-boolean -G: option specification: ' + arg)
+            arg = '-Dgraal.' + arg[len('-G:'):]
+        return arg
+    # add default graal.options.file and translate -G: options
+    return ['-Dgraal.options.file=graal.options'] + map(_translateGOption, args)
```
tkrodriguez: Shouldn't this be an absolute path into the suite?  Some of the mx commands change their directory before running things so it won't be found in this case.

dougxc: In addition, we should only add the system property to the command line if the file actually exists.

dougxc:
No need for a changelog entry as this is fixing a regression in expected behavior.

mx.graal-core/mx_graal_8.py
zapster: @tkrodriguez  right, suite path is probably better.
@dougxc ok, can do.

PR #85, opened 2016-02-26
Dump graph on failing assert

tkrodriguez:
try to reproduce the g1 assertion failure on jdk9


tkrodriguez:
I saw another failure with G1 barriers in JDK9 so I added some debug output.  I tried running the pull request repeatedly to reproduce it but without luck.  Hopefully it will fail for someone and we'll get some useful output.

PR #86, opened 2016-02-29
GraalCompiler: use more precise compilationUnitName if available.

zapster:
This makes compilation unit name more precise, e.g. for Truffle. The code for getting the JavaMethod is similar to the one used in the graph dumper. Is there a better way to do this other than looking in the debug context?

/cc @woess

### DIFF
--- a/graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/GraalCompiler.java
+++ b/graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/GraalCompiler.java
@@ -218,11 +218,11 @@ public class GraalCompiler {
             for (int i = 0; i < EmitLIRRepeatCount.getValue(); i++) {
                 SchedulePhase dummySchedule = new SchedulePhase();
                 dummySchedule.apply(graph);
-                emitLIR(backend, graph, stub, registerConfig, lirSuites);
+                emitLIR(backend, graph, stub, registerConfig, lirSuites, compilationResult);
             }
 
             LIRGenerationResult lirGen = null;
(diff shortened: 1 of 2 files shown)


tkrodriguez:
Isn't CompilationResult.getName the Truffle name of the compilation?

zapster:
Thanks @tkrodriguez. `CompilationResult.getName` is for sure better since it seems to be always available (not only in debug mode). PR updated.

graal/com.oracle.graal.compiler/src/com/oracle/graal/compiler/GraalCompiler.java
```diff
@@ -310,6 +305,17 @@ private static LIRGenerationResult emitLIR0(Backend backend, StructuredGraph gra
         }
     }
 
+    protected static <T extends CompilationResult> String getCompilationUnitName(StructuredGraph graph, T compilationResult) {
+        if (compilationResult != null && compilationResult.getName() != null) {
```
zapster: It seems that `CompilationResult.getName()` is always null for normal graal compilation ([see here](https://github.com/graalvm/graal-core/blob/6daee4240193a87f43f6b8586f2749e4bae9816e/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotGraalCompiler.java#L131)), but for truffle it is more precise than the `graph.method()` if it is available.

tkrodriguez: That's what I hoped.  Looks good.

PR #87, opened 2016-02-29
Add MethodProfilingPhase.

zapster:
The phase adds benchmark counters to the method entry (`METHOD_INVOCATION_COUNTER`) and loop headers (`METHOD_ITERATION_COUNTER`).

Usage Example:
```
mx dacapo -G:+LIRProfileMethods -G:+GenericDynamicCounters -XX:JVMCICounterSize=2000 sunflow -n 30
```


gilles-duboscq:
looks good

tkrodriguez:
How does this differ from ProfileCompiledMethodsPhase?

zapster:
@tkrodriguez Oh, forgot about that, thanks for bringing this back to mind. It looks like `ProfileCompiledMethodsPhase` is more precise. On the other hand it needs to build a schedule and a CFG which comes at a cost. But honestly, it is a debug facility so this might never be an issue. We can remove the `MethodProfilingPhase` if we see no other usecase.

PR #88, opened 2016-03-01
verify debug usage: check for parameterless toString() calls

davleopo:
cc @zapster

### DIFF
--- a/graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java
+++ b/graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java
@@ -60,8 +60,15 @@ public class VerifyDebugUsage extends VerifyPhase<PhaseContext> {
                                     String holder = m.getDeclaringClass().getName();
                                     if (holder.equals("Ljava/lang/StringBuilder;") || holder.equals("Ljava/lang/StringBuffer;")) {
                                         StackTraceElement e = graph.method().asStackTraceElement(invoke.bci());
-                                        throw new VerificationError(String.format("%s: parameter %d of call to %s appears to be a String concatenation expression.%n"
-                                                        + "    Use one of the multi-parameter Debug.log() methods or Debug.logv() instead.", e, argIdx, callee.format("%H.%n(%p)")));
+                                        throw new VerificationError(String.format("%s: parameter %d of call to %s appears to be a String concatenation expression.%n" +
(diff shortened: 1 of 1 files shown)


zapster:
Looks good. Btw, I noticed that we are only checking `Debug.log`. We should do the same for `Debug.logAndIndent` but I think this is out of the scope of this PR.

davleopo:
@zapster , yes maybe we should start checking all the other debug methods (not only the `log` calls) for invalid calls to `toString`. In think `Debug.log*`,`Debug.logAndIndent*` and also `Debug.dump*` are potential candidates for calls to `toString`

@dougxc what do you think?

graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java
```diff
@@ -60,8 +60,15 @@ protected boolean verify(StructuredGraph graph, PhaseContext context) {
...
-                                                        + "    Use one of the multi-parameter Debug.log() methods or Debug.logv() instead.", e, argIdx, callee.format("%H.%n(%p)")));
+                                        throw new VerificationError(String.format("%s: parameter %d of call to %s appears to be a String concatenation expression.%n" +
+                                                        "    Use one of the multi-parameter Debug.log() methods or Debug.logv() instead.", e, argIdx, callee.format("%H.%n(%p)")));
+                                    }
+                                    if (m.getSignature().getParameterCount(false) == 0 && m.getSignature().getReturnType(m.getDeclaringClass()).equals(
+                                                    context.getMetaAccess().lookupJavaType(String.class))) {
+                                        StackTraceElement e = graph.method().asStackTraceElement(invoke.bci());
+                                        throw new VerificationError(String.format("%s: parameter %d of call to %s appears to be a toString call without a parameter. %n " +
```
dougxc: The error message could be clearer. Something like:
`"%s: parameter %d of call to %s is a call to toString() which is redundant (the callee will do it) and forces unnecessary eager evaluation"`

dougxc:
Yes, I think we should broaden this checking to all relevant methods in `Debug`.

graal/com.oracle.graal.phases/src/com/oracle/graal/phases/verify/VerifyDebugUsage.java
davleopo: thanks, will do

davleopo:
ok, then I will prepare a new PR inlcuding all checks for the relevant methods in `Debug`.
Should I still update this PR or can we close it and open a new one including all changes?

PR #90, opened 2016-03-01
TraceRA: construct SSI after base pointer marking and clean-ups

zapster:
Currently mark base pointer phase is executed after SSI construction which increases runtime due to the increased number of operands. This change moves SSI construction after the marking phase to avoid this.
It also contains other minor clean-ups of debug code, options and visitor calls.

### DIFF
--- a/graal/com.oracle.graal.lir/src/com/oracle/graal/lir/alloc/trace/TraceRegisterAllocationPhase.java
+++ b/graal/com.oracle.graal.lir/src/com/oracle/graal/lir/alloc/trace/TraceRegisterAllocationPhase.java
@@ -48,7 +48,7 @@ import com.oracle.graal.lir.ssi.SSIUtil;
 import com.oracle.graal.lir.ssi.SSIVerifier;
 import com.oracle.graal.options.Option;
 import com.oracle.graal.options.OptionType;
-import com.oracle.graal.options.OptionValue;
+import com.oracle.graal.options.StableOptionValue;
 
 import jdk.vm.ci.code.TargetDescription;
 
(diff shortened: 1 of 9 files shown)


graal/com.oracle.graal.lir/src/com/oracle/graal/lir/alloc/trace/TraceRegisterAllocationPhase.java
```diff
@@ -61,13 +61,13 @@
...
+        public static final StableOptionValue<Boolean> TraceRAuseInterTraceHints = new StableOptionValue<>(true);
         @Option(help = "Use special allocator for trivial blocks.", type = OptionType.Debug)
-        public static final OptionValue<Boolean> TraceRAtrivialBlockAllocator = new OptionValue<>(true);
+        public static final StableOptionValue<Boolean> TraceRAtrivialBlockAllocator = new StableOptionValue<>(true);
         @Option(help = "Share information about spilled values to other traces.", type = OptionType.Debug)
-        public static final OptionValue<Boolean> TraceRAshareSpillInformation = new OptionValue<>(true);
+        public static final StableOptionValue<Boolean> TraceRAshareSpillInformation = new StableOptionValue<>(true);
         @Option(help = "Reuse spill slots for global move resolution cycle breaking.", type = OptionType.Debug)
-        public static final OptionValue<Boolean> TraceRAreuseStackSlotsForMoveResolutionCycleBreaking = new OptionValue<>(true);
+        public static final StableOptionValue<Boolean> TraceRAreuseStackSlotsForMoveResolutionCycleBreaking = new StableOptionValue<>(true);
```
rschatz: Any particular reason why these should be stable?

zapster: It saw calls to `OptionValue.getValue()` for (some of) those instances in a profile of a "hot" code fragment. I could cache the calls but in the end I think we do not need to set them at run time so making them stable seemed like a good decision.

PR #92, opened 2016-03-01
Introduce CheckedJavaType to avoid recomputing assumptions

axel22:
Introduce `CheckedJavaType` and remove duplicate creations of `Assumption` for exact types.
This refactoring should simplify the usage of assumptions and make it less error-prone.

Review by @thomaswue


thomaswue:
Subsumed by #101.

PR #94, opened 2016-03-02
Introduce new parser mode that forces explicit exception checks.

rschatz:
Previously, you could only specify `omitAllExceptionEdges`, but the reverse is also useful, for methods that should never deopt (e.g. in a static compiler setting, like it is used in the NFI).

The `CompiledNullPointerExceptionTest` is exercising this behavior.


graal/com.oracle.graal.compiler.test/src/com/oracle/graal/compiler/test/UnbalancedMonitorsTest.java
```diff
@@ -81,7 +81,6 @@ private void checkForBailout(String name) throws ClassNotFoundException {
         try {
             StructuredGraph graph = new StructuredGraph(method, AllowAssumptions.NO);
             GraphBuilderConfiguration graphBuilderConfig = GraphBuilderConfiguration.getEagerDefault(new Plugins(new InvocationPlugins(getMetaAccess())));
-            graphBuilderConfig = graphBuilderConfig.withOmitAllExceptionEdges(false);
```
rschatz: That's actually the default, so this line didn't do anything.

gilles-duboscq:
:+1:

PR #95, opened 2016-03-02
TraceRA: add faster SSI construction phase

zapster:
The FastSSIConstruction phase uses the same implementation as the LinearScanLifetimeAnalysisPhase which is faster than the naive original implementation.


graal/com.oracle.graal.lir/src/com/oracle/graal/lir/ssi/FastSSIConstructionPhase.java
```diff
@@ -0,0 +1,399 @@
...
+
+                            /*
+                             * liveOut(block) is the union of liveIn(sux), for successors sux of
+                             * block.
+                             */
+                            int n = block.getSuccessorCount();
+                            if (n > 0) {
+                                liveOut.clear();
+                                // block has successors
+                                if (n > 0) {
```
gilles-duboscq: We just checked `n > 0` 3 lines above. Is this a `ConditionalElimination` challenge? :wink:

zapster: No, it is a challenge for the reviewer ;). Thanks for catching!

PR #99, opened 2016-03-04
Fix resetting the marker for the first frame in frame traversal was done wrongly.

chumer:


woess:
Thanks for the fix. Merged. This would deserve a regression test, though.

PR #101, opened 2016-03-07
Introduce TypeReference class.

thomaswue:


thomaswue:
@tkrodriguez @dougxc

thomaswue:
@davleopo

dougxc:
For a commit this size, a non-empty description can help expedite reviewing the code.

graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/type/TypeReference.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    private static ResolvedJavaType filterInterfaceTypesOut(ResolvedJavaType type) {
+        if (type != null) {
+            if (type.isArray()) {
+                ResolvedJavaType componentType = filterInterfaceTypesOut(type.getComponentType());
+                if (componentType != null) {
+                    return componentType.getArrayClass();
+                }
+                return type.getSuperclass().getArrayClass();
```
axel22: Nitpick: It seems that this line will always return an `Object[]` type. Maybe it would be more readable to write `Object[].class`.

thomaswue:
@davleopo

graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/type/TypeReference.java
```diff
@@ -0,0 +1,155 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.graal.compiler.common.type;
+
+import jdk.vm.ci.meta.Assumptions;
+import jdk.vm.ci.meta.ResolvedJavaType;
+
+/**
+ * This class represents a reference to a Java type and whether this reference is refers only to the
```
davleopo: I think this should be "is referring" or "this reference refers "

davleopo:
@thomaswue maybe the removal of the old `ConditionalEliminationPhase` would be worth a separate commit/PR as it is not directly related to the new typereference?

graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/type/TypeReference.java
thomaswue: OK, fixed.

thomaswue:
@davleopo Agreed. Will push separately.

graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/type/TypeReference.java
thomaswue: I added a comment for readability.

axel22:
It might be useful to delete the `t-refactor-branch` at this point.

PR #102, opened 2016-03-07
Fix double-checked locking pattern in OptimizedCallTarget

eregon:
* Only publish the OptimizedCallTarget as initialized when it is.
* `intialized` needs to be volatile or the store to it might move above the actual initialization.


chumer:
Just moving the initialized=true after the call should be enough,  no? We don't need volatile here... I think.

chrisseaton:
But if the write to `initialized` isn't volatile it can float all the way back up couldn't it?

woess:
@chrisseaton **very** unlikely here, but theoretically, yes

smarr:
The `volatile` seems to be necessary since access to the field in the checks isn't synchronized.

eregon:
We do. The compiler or the CPU (not x86 or SPARC TSO) can move that write above.

PR #104, opened 2016-03-07
SPARC: Bail out when branch displacements exceed 19 bit

sanzinger:
This change lets graal bail out gracefully when branch displacements do not fit in 19 bit signed field.
Remove redundant SPARCAssembler.{bpcc,fbpcc,bicc,...} variants and use the  SPARCAssembler.{BPR,BPICC,FBPCC,...} variants instead, which do the check and bail out.

A proper fix would require some kind of trampolines or some more complex handling in the two-pass assembly. Since this problem occurs only on very huge methods (>1MB assembler code) we don't consider this as a major problem since the VM can fall back into the Interpreter.

The issue was reported in https://bugs.openjdk.java.net/browse/GRAAL-60


christhalinger:
Looks good.

graal/com.oracle.graal.asm.sparc/src/com/oracle/graal/asm/sparc/SPARCAssembler.java
```diff
@@ -919,14 +920,31 @@ public int getValue() {
...
 
+        protected final boolean bailoutOnOverflow;
+
+        public BitSpec(boolean bailoutOnOverflow) {
+            super();
+            this.bailoutOnOverflow = bailoutOnOverflow;
+        }
+
+        protected void checkValue(int value) {
+            if (bailoutOnOverflow) {
```
tkrodriguez: Why is this conditional?  Shouldn't it always be an error?

graal/com.oracle.graal.asm.sparc/src/com/oracle/graal/asm/sparc/SPARCAssembler.java
```diff
@@ -919,14 +920,31 @@ public int getValue() {
...
+                if (!valueFits(value)) {
+                    throw new BailoutException("Value 0x%x for field %s does not fit.", value, this);
```
tkrodriguez: It might be nicer to have a more specific exception thrown from the assembler.  The assembler driver would be responsible for catching that and converting it into the appropriate action.

tkrodriguez:
I would usually expect errors of this kind to be handled above the level of the assembler itself.   Don't we have to have displacement reachability checks in other places?

graal/com.oracle.graal.asm.sparc/src/com/oracle/graal/asm/sparc/SPARCAssembler.java
sanzinger: Yes, it would be nice to have it always checked. Just left it out due to performance concerns (HS does such sort of checks only in debug builds) 
I'm taking your suggestion and check, what impact it has, when we do the check always.

graal/com.oracle.graal.asm.sparc/src/com/oracle/graal/asm/sparc/SPARCAssembler.java
```diff
@@ -998,35 +1020,49 @@ public boolean valueFits(int value) {
...
             int r = right.getBits(word);
+            if (signExt) {
+                l = signExtend(l, leftWidth);
+            }
             return l << rightWidth | r;
         }
 
         @Override
         public int setBits(int word, int value) {
+            checkValue(value);
```
dougxc: Do we really need the check here given that the setting the bits in the composite parts will also perform checks?

sanzinger: Yes, this is redundant. We should just set the left bits as sign extended if necessary. I'll fix.

sanzinger:
@tkrodriguez 
> I would usually expect errors of this kind to be handled above the level of the assembler itself. Don't we have to have displacement reachability checks in other places?
Yes this is correct, but I wanted to have a place, where no user of a branc instruction can "forget" such checks, thus I've put it in the lowest level. But I agree, it is to low.

I'm now putting the check into ControlTransfer.setDisp(int, int), where the displacements for all control-transfer instructions are set. There we can just have a check whether the value fits or not, and then do a BailOut. This is the best place in my opinion. In any higher level (outside the assembler) it doesn't make sense since all code outside the assembler deals with Labels instead of concrete displacements.

The exception looks as follows, (unittest created for this):
```
jdk.vm.ci.code.BailoutException: Too large displacement 0x2ff in field CompositeBitSpec[d10Hi [20:19], d10Lo [12:5]] of instruction SPARCAssembler$CBCond[op: BranchOp]
        at com.oracle.graal.asm.sparc.SPARCAssembler$ControlTransferOp.setDisp(SPARCAssembler.java:1294)
        at com.oracle.graal.asm.sparc.test.SPARCAssemblerTest.testControlTransferInvalidDisp(SPARCAssemblerTest.java:118)
```

tkrodriguez:
That looks better.  One way or another there has to be some non-assert logic that verifies that branches support their displacements.  In HotSpot that's handled above the level of the assembler.  Putting it at the binding of the displacements is a good compromise for us.  Looks good.

graal/com.oracle.graal.asm.sparc/src/com/oracle/graal/asm/sparc/SPARCAssembler.java
```diff
@@ -923,10 +924,27 @@ public int getValue() {
...
+        protected final boolean signExtend;
+
+        public BitSpec(boolean signExtend) {
+            super();
+            this.signExtend = signExtend;
+        }
+
+        protected void checkValue(int value) {
+            if (!valueFits(value)) {
+                throw new InternalError(String.format("Value 0x%x for field %s does not fit.", value, this));
```
dougxc: Why `InternalError`? This is still a bailout condition caused by compiler input as far as I can tell.

tkrodriguez: Shouldn't this revert back to being an assert?  I assume the setDisp check should ensure this is all done correctly.

dougxc: Ah yes, I missed that the _real_ check is done in `setDisp`.

sanzinger: But you're correct, we can turn the range check into an assert again. The only unbounded values in the assembler are the branch displacements. All other values either origin from Enums (Opcodes and flags) or are immediate fields, which are checked outside the assembler anyway. Thus, we can make an assertion out of it.

tkrodriguez:
looks good to me.

PR #106, opened 2016-03-08
Try out javac-daemon

tkrodriguez:
test out of mx javac daemon support


smarr:
@tkrodriguez some of you PRs looks like they are just tests. Do you know that you can enable Travis for your on repository? It will then run for all your pushes, so, you don't need these 'test pull requests'.

tkrodriguez:
I wondered about that.  It will run it for everything though right?  I guess I can just enable travis when I want some testing done.  I'll do that instead.

smarr:
Normally it will run for every push, yes.
If you really care about skipping Travis you can mark a commit with `[skip ci]`, I think.
Or abort it via the button in the Travis UI.

tkrodriguez:
I think I can enable and disable it as needed.  Thanks for the tip.

PR #107, opened 2016-03-08
Remove ResolvedJavaType.asExactType

tkrodriguez:


gilles-duboscq:
:+1:

tkrodriguez:
Thomas actually purged all uses in his rewrite so this just ends up being a JVMCI import bump.

PR #108, opened 2016-03-09
Simplify instanceof node.

thomaswue:


axel22:
LGTM to me overall. Please sync with the `master` branch so that we can merge this PR.

thomaswue:
@axel22 I've improved the code to also handle the case where two incompatible interface types with a common super interface type are involved. I've extended the tests to include this corner case.

thomaswue:
@axel22 See commit https://github.com/graalvm/graal-core/commit/ada793a36a26c5140ec79afe1a1c011fc3d9c87d

PR #100, opened 2016-03-04
Cleanup and refactor OSR; Move to local counters instead of a global counter in CallTarget; Implemented revirtualization of local variables usable by internal API. Added OSR test suite.

chumer:
Cleanup and refactor OSR; 
Move to local counters instead of a global counter in CallTarget; 
Implemented revirtualization of local variables usable by internal API. 
Added OSR test suite.

This is quite a serious change of the OSR mechanism.
@woess  please review and verify that it works with JS
@chrisseaton please verify that this works Ruby

Please note that the default OSR threshold is quite high atm.

I will verify/verified it for Fastr (cc @lukasstadler).

PR #109, opened 2016-03-09
gate: make truffle compilation errors in unittests fatal.

zapster:
Truffle compilation exceptions are currently not fatal for gate unittests.

### DIFF
--- a/mx.graal-core/mx_graal_8.py
+++ b/mx.graal-core/mx_graal_8.py
@@ -315,7 +315,7 @@ def compiler_gate_runner(suites, unit_test_runs, bootstrap_tests, tasks, extraVM
 
 
 graal_unit_test_runs = [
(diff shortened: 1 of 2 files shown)


gilles-duboscq:
:+1:

woess:
:+1:

tkrodriguez:
We recently made ExitVMOnException true when asserts are enabled in #82.  Maybe something similar should be done for TruffleCompilationExceptionsAreFatal?

zapster:
IMHO this would be nice, although I'm not sure about the implications downstream. I'll prepare a PR for discussion.

Also, since there are currently exceptions in Truffle unittests (which are probably fixed by #100, graalvm/truffle#112 and graalvm/truffle#113) I wanted to make them fatal before we start running them as part of the graal-core Travis checks (which I believe is the cast once we merge #100 and therefore update the Truffle import version). Silently failing tests are not really useful.

PR #110, opened 2016-03-09
Remove `CStringNode`, use `ConstantNode` with `CStringConstant` instead.

rschatz:
Just some cleanup opportunity I stumbled over while working on the exception stubs. That way, the constant optimizations by the backend and register allocator will also apply to raw c string constants.

This passes all unit tests on AMD64 and SPARC locally. I hope I didn't break AArch64, but since it's only a change to the interface of the backend, not to the implementation, my chances are good ;)

/cc @sanzinger


christhalinger:
Do you guys have access to an ARMv8 board?

dougxc:
Not where I am.

dougxc:
lgtm

zapster:
@christhalinger no :|

tkrodriguez:
looks good

PR #113, opened 2016-03-09
Enable TruffleCompilationExceptionsAreFatal when asserts are enabled.

zapster:
Same idea as #82 but for `TruffleCompilationExceptionsAreFatal`. I'm not completely sure about the downstream implications so reviews from Truffle folks would be great, maybe @woess @chumer @chrisseaton?

Note that this supersedes #109.

### DIFF
--- a/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
+++ b/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
@@ -397,9 +397,16 @@ public class OptimizedCallTarget extends InstalledCode implements RootCallTarget
             if (TruffleCompilationExceptionsAreThrown.getValue()) {
                 throw new OptimizationFailedException(t, this);
             }
-            if (TruffleCompilationExceptionsArePrinted.getValue() || TruffleCompilationExceptionsAreFatal.getValue()) {
+            /*
+             * Automatically enable TruffleCompilationExceptionsAreFatal when asserts are enabled
+             * but respect TruffleCompilationExceptionsAreFatal if it's been explicitly set.
+             */
+            boolean truffleCompilationExceptionsAreFatal = TruffleCompilationExceptionsAreFatal.getValue();
+            assert TruffleCompilationExceptionsAreFatal.hasBeenSet() || (truffleCompilationExceptionsAreFatal = true) == true;
+
+            if (TruffleCompilationExceptionsArePrinted.getValue() || truffleCompilationExceptionsAreFatal) {
                 printException(t);
-                if (TruffleCompilationExceptionsAreFatal.getValue()) {
+                if (truffleCompilationExceptionsAreFatal) {
                     System.exit(-1);
                 }
             }
--- a/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/TruffleCompilerOptions.java
+++ b/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/TruffleCompilerOptions.java
@@ -136,7 +136,7 @@ public class TruffleCompilerOptions {
     public static final OptionValue<Boolean> TruffleCompilationExceptionsAreFatal = new OptionValue<>(false);
 
     @Option(help = "Prints the exception stack trace for compilation exceptions", type = OptionType.Debug)
-    public static final OptionValue<Boolean> TruffleCompilationExceptionsArePrinted = new OptionValue<>(false);
+    public static final OptionValue<Boolean> TruffleCompilationExceptionsArePrinted = new OptionValue<>(true);
 
     @Option(help = "Treat compilation exceptions as thrown runtime exceptions", type = OptionType.Debug)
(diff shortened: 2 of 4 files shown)


tkrodriguez:
Looks good to me.  Maybe we direct support for this in the option machinery would be useful.

woess:
We should probably enable TruffleCompilationExceptionsArePrinted by default as well.
It's unclear to me whether we want this exit-vm-on-exception behavior depend on `-esa`, but I'm ok with it.

zapster:
Thanks for the reviews!
@tkrodriguez Yes, if we find another usecase we should outsource the mechanism.
@woess "by default" as in "unconditionally"? About the exit behavior: I believe the exit-on-error is the safer default. For example graalvm/truffle#113 was not discovered for some time because of the current setting.

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
```diff
@@ -397,9 +397,16 @@ public void notifyCompilationFailed(Throwable t) {
             if (TruffleCompilationExceptionsAreThrown.getValue()) {
                 throw new OptimizationFailedException(t, this);
             }
-            if (TruffleCompilationExceptionsArePrinted.getValue() || TruffleCompilationExceptionsAreFatal.getValue()) {
+            /*
+             * Automatically enable TruffleCompilationExceptionsAreFatal when asserts are enabled
+             * but respect TruffleCompilationExceptionsAreFatal if it's been explicitly set.
+             */
+            boolean truffleCompilationExceptionsAreFatal = TruffleCompilationExceptionsAreFatal.getValue();
+            assert TruffleCompilationExceptionsAreFatal.hasBeenSet() || (truffleCompilationExceptionsAreFatal = true) == true;
```
woess: nitpicking: the `== true` part looks redundant

woess:
as in `TruffleCompilationExceptionsArePrinted = new OptionValue<>(true);`. could you add it to this PR?

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
zapster: I guess you are right, just copy pasted from #82. I think I'll follow @tkrodriguez' advice and move the functionality to `OptionValue`.

zapster:
@woess can do but I think this means we need to reevaluate the PR. Removing the `accept` label...

graal/com.oracle.graal.options/src/com/oracle/graal/options/OptionValue.java
```diff
@@ -331,6 +331,17 @@ public void setValue(Object v) {
     }
 
     /**
+     * Returns the options {@linkplain OptionValue#getValue() value} if it has been
+     * {@linkplain OptionValue#hasBeenSet() explicitly set} or otherwise {@code true} if assertions
+     * are enabled.
+     */
+    public static boolean assertsEnabledOrValueIfSet(OptionValue<Boolean> booleanOptionValue) {
```
zapster: Not quite sure about the method name (and javadoc). Any suggestions? (/cc @dougxc)

dougxc: The problem with checking whether assertions are enabled is that they can be disabled/enabled per class (e.g., `-ea[:<packagename>...|:<classname>]` and `-da[:<packagename>...|:<classname>]`) so really this method is asking whether the boolean option is true or if assertions are enabled for `OptionValue`. In practice, almost the assertion status for the caller and `OptionValue` will be the same but it could be mightily confusing when they aren't. As such, we should not have a static helper like this.

zapster: Oh, right. I'll revert this change.Thanks!

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
woess: sounds like a good idea!

woess:
Thanks! Looks good to me.

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
tkrodriguez: And I was copy pasting from somewhere else.  I think the == true is intended to make it clear the assignment is intentional.

tkrodriguez:
According to the email I got it looks like there was a little attempt to integrate this into OptionValue but it seems to have disappeared from the PR.  Anyway, having the machinery built in would be tricky because it would move around which class would trigger the assert.  I think we could have something like AssertionOptionValue which is always a boolean false by default that defaults to true if assertions are enabled in that class.  We already have the problem that fine grained option enabling won't really necessarily turn it on so it wouldn't change that.

PR #115, opened 2016-03-09
Correctly check for progress when removing phi from list during peel

tkrodriguez:
Removing the element from the list changes the progress conditions so the previous fix in #114 will only go through the list once.  Use a progress flag instead.


dougxc:
This also is easier to follow.

tkrodriguez:
Agreed.  Thanks!

PR #120, opened 2016-03-10
Try to fix WriteBarrierAdditionTest on JDK9

tkrodriguez:
The debug output I added finally triggered and it looks like we're terminating the parse with a deopt so we have fewer than expected barriers.  From https://travis-ci.org/graalvm/graal-core/builds/115086162

 `0|StartNode    (0)`
  `1|NewInstanceNode    (4)`
  `2|NewInstanceNode    (3)`
 ` 3|CompressionNode    (2)`
  `4|OffsetAddressNode    (3)`
 ` 5|G1PreWriteBarrier    (0)`
  `6|WriteNode    (0)`
 ` 7|G1PostWriteBarrier    (0)`
  `8|DeoptimizeNode    (0)`
 `expected:<5> but was:<2>`

It looks like we're deoptimizing before the get call.  It's possible this code hasn't been exercised at the point we compile it.  I've added an invoke of the method to ensure everything has been resolved but I we won't really know for it's fixed.

### DIFF
--- a/graal/com.oracle.graal.hotspot.test/src/com/oracle/graal/hotspot/test/WriteBarrierAdditionTest.java
+++ b/graal/com.oracle.graal.hotspot.test/src/com/oracle/graal/hotspot/test/WriteBarrierAdditionTest.java
@@ -147,6 +147,8 @@ public class WriteBarrierAdditionTest extends HotSpotGraalCompilerTest {
      */
     @Test
     public void test4() throws Exception {
+        // Exercise the snippet before compiling it to ensure that everything is resolved.
+        test4Snippet();
         testHelper("test4Snippet", config.useG1GC ? 5 : 2);
     }
 


zapster:
closed/reopened to retrigger Travis

gilles-duboscq:
Does the restart button from Travis not work for this use-case?

zapster:
@gilles-duboscq It works but it deletes the old log. I wanted to keep it for reference so I did the close/open trick.

gilles-duboscq:
Ok, good to know

rschatz:
lgtm

dougxc:
What exactly could it be deoptimizing on? If it's a case of something not having been executed, then surely this command should trigger the issue (but obviously doesn't):
```
dsimon@kurz ~/g/graal-core> mx --java-home ~/ojn-graal-jdk9/build/macosx-x86_64-normal-server-release/images/jdk unittest -XX:+UseG1GC WriteBarrierAdditionTest#test4
MxJUnitCore
JUnit version 4.11
.
Time: 0.343

OK (1 test)
```

rschatz:
I'm not sure I fully understand the reason for the failure either, but since we can't reproduce it reliably, I don't see what else we can do other than just try it out and see whether it happens again...

tkrodriguez:
I'm just guessing that we put in a deopt before the Reference.get call for some reason.  The most obvious one is that we haven't ensured that path has ever been executed so maybe something isn't resolved when we get there.  I could add more diagnostic checking to figure out why it doesn't get inlined instead of attempting a blind fix.   At least printing out the deopt reason and maybe the frame state associated with it might tell us something.

PR #124, opened 2016-03-11
PE: @ExplodeLoop(merge=false) should use LoopExplosionKind.FULL_UNROLL

woess:
This affects, e.g., code of the form:
```java
@ExplodeLoop
public void executeVoid(VirtualFrame frame) {
    int n = statements.length;
    for (int i = 0; i < n; i++) {
        if (/* some condition */) {
            statements[i].executeVoid(frame);
        }
    }
}
```
Graal simplifies the Merge |> LoopEnd to 2 LoopEnds and this leads to an exponential number of exploded loop iterations with `LoopExplosionKind.FULL_EXPLODE`.
I don't think this behavior is intentional or desired. So the solution is to change it to `FULL_UNROLL`, which is what I would expect to happen here (i.e., loop unrolled `n` times).
Please let me know if you do have a case where you really want the full explosion.

### DIFF
--- a/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/SimplifyingGraphDecoder.java
+++ b/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/SimplifyingGraphDecoder.java
@@ -199,18 +199,18 @@ public class SimplifyingGraphDecoder extends GraphDecoder {
                      */
                     canonical = new CanonicalizeToNullNode(node.stamp);
                 }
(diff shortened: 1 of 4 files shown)


lukasstadler:
looks good.

christianwimmer:
As long as no Truffle language complains about the change, I'm fine with it.  Originally, I made FULL_EXPLODE the default because it improved JavaScript performance.

If any language needs the old semantics, the best solution is to expose all levels to the @ExplodeLoop annotation, i.e., change the parameter from a boolean to an enum.

PR #125, opened 2016-03-11
Update jvmci import: Change property name jvmci.compiler -> jvmci.Compiler

rschatz:
This recent jdk9 change changes the property name to select the compiler to upper case:
http://hg.openjdk.java.net/jdk9/hs-comp/hotspot/rev/74522870dbde

This pull request updates the mx scripts to deal with that, making us compatible with the current jdk9 version again. It also updates the jvmci import to include a backport of the jdk9 change to jvmci-8, to keep the API in sync.


tkrodriguez:
This has the side effect that you have to specify -Djava.Compiler=graal on the command line in jdk8.  Was that an intentional part of these changes?  JDK9 requires that it's always specified but that's never been the case for 8.  I think we should restore the init_system_properties part of the JVMCI side of these changes.

dougxc:
That explains why a number of downstream users are seeing:
```
jdk.vm.ci.common.JVMCIError: JVMCI system compiler 'jdk.vm.ci.hotspot.HotSpotJVMCICompilerConfig$DummyCompilerFactory' is not a Graal compiler.
	at com.oracle.graal.truffle.hotspot.HotSpotTruffleRuntimeAccess$LazySystemGraalRuntime.getCompiler(HotSpotTruffleRuntimeAccess.java:118)
```

dougxc:
I'm restoring `init_system_properties` in jvmci-8 now.

rschatz:
Sorry, I missed that side effect. I just backported the change from jvmci-9 and didn't realize we have different behaviour in 8 and 9...

tkrodriguez:
Yeah I found the mechanism somewhat too subtle to see at a glance.  A comment in the code that it configures the default compiler would be helpful.

PR #126, opened 2016-03-14
Add JavaInteropSpeedBench.

zapster:
JMH micro benchmark derived from `JavaInteropSpeedTest` (cf graalvm/truffle#112). The test can be executed as follows:
```
mx microbench -- com.oracle.graal.truffle.bench.interop.JavaInteropSpeedBench
```


graal/com.oracle.graal.truffle.bench/src/com/oracle/graal/truffle/bench/interop/JavaInteropSpeedBench.java
```diff
@@ -0,0 +1,95 @@
...
+@Fork(1)
+@State(Scope.Thread)
+public class JavaInteropSpeedBench {
+    private static final int REPEAT = 10000;
+    private static final long SEED = 42;
+    private static int[] arr;
+
+    @Setup
+    public static void beforeTesting() {
+        InstrumentationTestMode.set(true);
```
woess: I believe we don't need `InstrumentationTestMode.set` anymore. Which means we should also remove it from the tests in the truffle repo. Maybe @chumer or @mlvdv can comment on that?

graal/com.oracle.graal.truffle.bench/src/com/oracle/graal/truffle/bench/interop/JavaInteropSpeedBench.java
```diff
@@ -0,0 +1,95 @@
...
+    @Benchmark
+    public int doMinMaxInJava() {
+        int max = 0;
+        for (int i = 0; i < arr.length; i++) {
+            max = Math.max(arr[i], max);
+        }
+        return max;
+    }
+
+    private static final TruffleObject TRUFFLE_MAX = JavaInterop.asTruffleFunction(IntBinaryOperation.class, new IntBinaryOperation() {
```
woess: I'd suggest to use `java.lang.function.IntBinaryOperator` here and drop the `IntBinaryOperation` class.

zapster:
Update: I've removed `InstrumentationTestMode` since we think we don't need it anymore (true? @chumer or @mlvdv) and it contained compilation errors.

woess:
looks good to me, but waiting for @jtulach's approval

chumer:
@zapster if it runs for you without InstrumentationTestMode you should be fine.

zapster:
@chumer thanks for the info

PR #127, opened 2016-03-14
generated exception stubs

rschatz:
Use Graal-generated stubs instead of handwritten C stubs for throwing NullPointerException and ArrayIndexOutOfBoundsException.

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.graal.hotspot.stubs;
+
+import com.oracle.graal.hotspot.HotSpotForeignCallLinkage;
+import com.oracle.graal.hotspot.meta.HotSpotProviders;
+import com.oracle.graal.replacements.Snippet;
+import com.oracle.graal.replacements.Snippet.ConstantParameter;
+
+import jdk.vm.ci.code.Register;
+import jdk.vm.ci.common.JVMCIError;
+
+/**
+ * Stub to allocate a {@link NullPointerException} thrown by a bytecode.
+ */
+public class NullPointerExceptionStub extends CreateExceptionStub {
+
+    public NullPointerExceptionStub(HotSpotProviders providers, HotSpotForeignCallLinkage linkage) {
+        super("createNullPointerException", providers, linkage);
+    }
+
+    @Override
+    protected Object getConstantParameterValue(int index, String name) {
+        JVMCIError.guarantee(index == 0, "unknown parameter %s at index %d", name, index);
+        return providers.getRegisters().getThreadRegister();
+    }
+
+    @Snippet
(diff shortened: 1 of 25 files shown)


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
```diff
@@ -0,0 +1,54 @@
...
+    public NullPointerExceptionStub(HotSpotProviders providers, HotSpotForeignCallLinkage linkage) {
+        super("createNullPointerException", providers, linkage);
+    }
+
+    @Override
+    protected Object getConstantParameterValue(int index, String name) {
+        if (index == 0) {
+            return providers.getRegisters().getThreadRegister();
+        } else {
+            return null;
```
tkrodriguez: this case should probably be an assert or an explicit throw

tkrodriguez:
After seeing all the changes required on both sides and the code required to construct the right messages I'm wondering if we should just be providing the same entry points that C1 has for throwing implicit exceptions in c1_Runtime1.cpp from JVMCIRuntime.  The set of exceptions that have to be handled in this way is limited I think since they are only required for exceptions which are thrown implicitly.  I guess we could have a case where we wanted stubs for some other case.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/OutOfBoundsExceptionStub.java
```diff
@@ -0,0 +1,86 @@
...
+            ptr = ptr.subtract(1);
+            ptr.writeByte(0, (byte) ('0' + digit));
+        } while (number > 0);
+
+        if (idx < 0) {
+            ptr = ptr.subtract(1);
+            ptr.writeByte(0, (byte) '-');
+        }
+
+        return createException(threadRegister, ArrayIndexOutOfBoundsException.class, ptr);
```
tkrodriguez: Don't you need to null terminate the string?

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
dougxc: Yes, the `index` should be asserted to be 0 before its use.

christhalinger: Why exactly?  The other stubs have a switch-case and don't assert in the default case.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/OutOfBoundsExceptionStub.java
dougxc: He's building the string backwards and starts with the null terminator.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
dougxc: Depends on which other stubs are you looking at. Both `ExceptionHandlerStub.getConstantParameterValue` and `UnwindExceptionToCallerStub.getConstantParameterValue` use assertions. These however should be `JVMCIError.guarantee` - there's no performance cost to unconditionally doing the check.

christhalinger: True.  I left a comment in `OutOfBoundsExceptionStub`.

tkrodriguez: These might make more sense if they were just @Fold methods.  Usually parameters are used to provide different variations but that's not happening here.  These are all singletons.

rschatz:
re "just export all exception throwing routines C1 has available":
We would be introducing a lot of code that's never exercised because Graal doesn't use it yet. And on the other hand, we'd need to be very sure that we don't miss anything. The solution with generating the stubs ourselves is more future-proof, we can always just create another stub.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/stubs/NullPointerExceptionStub.java
rschatz: These `@Fold` methods would not be static, so we need to get something in there, either the `HotSpotProviders`, or the `this` object. And when we're passing in something, we might as well directly pass the `Register` object.

For the others (in OutOfBoundsExceptionStub), i agree, an `@Fold` method might be better.

rschatz:
New version addressing the comments raised so far:
- wrong index in `getConstantParameterValue` now throws an exception
- removed `bufferSize` constant parameter, since it doesn't need access to `providers` (the suggestion was to use an `@Fold` method, but we actually don't need even that)

I've also included a `SymbolPointer` word-type, analogous to the existing `KlassPointer` and `MethodPointer` types.

What I haven't addressed yet is that we need variants of `throw_and_post_jvmti_exception` for exception messages that contain a `Klass::internal_name()` (`ArrayStoreException` and `ClassCastException`). Since those two are currently not used in Graal, I'd like to merge this PR now, and do a separate PR for supporting the others.

tkrodriguez:
I meant to comment last night that it looks good to me, but forgot to, so just for the record, lgtm.

PR #129, opened 2016-03-15
TraceRA: outsource tracebuilding

zapster:
Trace building was previously called from within `TraceRegisterAllocation`. Doing this as separate phase is more modular.


gilles-duboscq:
:+1:

PR #130, opened 2016-03-15
TraceRA: remove old SSIConstructionPhase and SSI Linear Scan

zapster:
- The `SSIConstructionPhase` has been superseded by a faster variant (see #95).
- The SSI linear scan was just a proof-of-concept implementation which is no longer needed.


rschatz:
Lgtm.

In case anyone else is looking at this: The total diff in this PR looks a lot more complicated than it actually is because Github messes up the display of a delete + rename combination. Look at the individual commits to see what really changed.

PR #131, opened 2016-03-15
Use unsigned shift when tracking locals changed in loops

tkrodriguez:
@christhalinger tripped across this while running runescape.


christhalinger:
Looks good.

PR #133, opened 2016-03-16
Use the microbench facilities provided by mx

zapster:


zapster:
This adopts graalvm/mx#85.

PR #135, opened 2016-03-16
Using TruffleServices instead of direct access to Accessor

jtulach:
This pull request contains changes in graal-core to use work done on
https://github.com/jtulach/truffle/tree/TruffleServices
branch. The changes in Truffle repository introduce new module `com.oracle.truffle.api.boot` - the idea is that at the end, this will be the only module on bootclasspath - the rest of the Truffle could stay on application classpath.

The module introduces `TruffleServices` class that shall include various methods returning *services* useful for the Truffle API. Right now it only provides the LoopCountSupport - that allows us to remove the dependency of Graal on `Accessor` class of Truffle.

I am not very happy with [current state](https://github.com/jtulach/truffle/commit/84f6950882a7d435737d09970a32dd6ee3765923), but at least it sketches the plan I am trying to follow. Comments welcomed.

### DIFF
--- a/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTVMCI.java
+++ b/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTVMCI.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2014, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.graal.truffle;
+
+import com.oracle.truffle.api.CallTarget;
+import com.oracle.truffle.api.TruffleLanguage;
+import com.oracle.truffle.api.impl.TVMCI;
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.RootNode;
+
+final class GraalTVMCI extends TVMCI {
+
+    @Override
+    public void onLoopCount(Node source, int count) {
+        Node node = source;
+        Node parentNode = source != null ? source.getParent() : null;
+        while (node != null) {
+            if (node instanceof OptimizedOSRLoopNode) {
+                ((OptimizedOSRLoopNode) node).reportChildLoopCount(count);
+            }
+            parentNode = node;
+            node = node.getParent();
+        }
+        if (parentNode != null && parentNode instanceof RootNode) {
+            CallTarget target = ((RootNode) parentNode).getCallTarget();
+            if (target instanceof OptimizedCallTarget) {
+                ((OptimizedCallTarget) target).onLoopCount(count);
+            }
+        }
+    }
+
+    @SuppressWarnings("rawtypes")
+    Class<? extends TruffleLanguage> findLanguage(RootNode root) {
+        return super.findLanguageClass(root);
+    }
+
+    void onFirstExecution(OptimizedCallTarget callTarget) {
+        super.onFirstExecution(callTarget.getRootNode());
+    }
+
+}
--- a/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
+++ b/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
@@ -68,6 +68,7 @@ import com.oracle.truffle.api.frame.FrameInstance;
 import com.oracle.truffle.api.frame.FrameInstanceVisitor;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.impl.TVMCI;
 import com.oracle.truffle.api.nodes.DirectCallNode;
 import com.oracle.truffle.api.nodes.IndirectCallNode;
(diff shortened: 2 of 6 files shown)


jtulach:
All checks have passed. The approach seems to work OK. We can discuss packaging, naming, etc. I'll start a review for the Truffle changes, so people can see the actual diff: https://github.com/graalvm/truffle/pull/139

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
```diff
@@ -84,7 +87,8 @@
 import jdk.vm.ci.meta.ResolvedJavaMethod;
 import jdk.vm.ci.services.Services;
 
-public abstract class GraalTruffleRuntime implements TruffleRuntime {
+public abstract class GraalTruffleRuntime extends TruffleServices implements TruffleRuntime {
+    static TruffleInfo INFO;
```
gilles-duboscq: Why do we use a static field here? That seems kind of dangerous.
Can't we have it as an instance field? `OptimizedCallTarget` has access to its associated `GraalTruffleRuntime` and in `OptimizedOSRLoopNode` since we already seem to be assuming a `GraalTruffleRuntime` we can just use `Truffle.getRuntime()` and cast

chumer: +1

jtulach: Thanks. I haven't noticed the `runtime` field in `OptimizedCallTarget` - in fact I wasn't expecting one as it needlessly contributes to the memory consumption in real system. Having multiple `GraalTruffleRuntime` instances is only useful in unit testing, I think.

Btw. the code in `OptimizedOSRLoopNode` is the typical mess I would expect in `OptimizedCallTarget` as well.

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedOSRLoopNode.java
```diff
@@ -211,12 +211,13 @@ public void run() {
...
+    @SuppressWarnings({"rawtypes", "unchecked"})
     private OSRRootNode createRootNodeImpl(RootNode root, Class<? extends VirtualFrame> frameClass) {
-        @SuppressWarnings("rawtypes")
-        Class<? extends TruffleLanguage> truffleLanguage;
+        Class truffleLanguage;
         FrameDescriptor frameDescriptor;
         if (root != null) {
-            truffleLanguage = OptimizedCallTarget.ACCESSOR.findLanguage(root);
+            GraalTruffleRuntime runtime = (GraalTruffleRuntime) Truffle.getRuntime();
+            truffleLanguage = runtime.truffleInfo().findLanguage(root);
```
jtulach: Shouldn't the `OptimizedOSRLoopNode` have a pointer to `runtime` just like `OptimizedCallTarget` has to avoid this singleton access+cast? Or vice-versa: change `OptimizedCallTarget` to access this singleton as well?

chumer: I think OptimizedCallTarget should access the singleton as well. It only happens for compilation events and there are actually not so many of them.

woess: I'm also don't think we need a `runtime` instance field in OptimizedCallTarget.

jtulach: I removed the field in most recent diff.

jtulach:
Both #135 and graalvm/truffle#139 are ready to be merged in my opinion. Please review.

chumer:
I still don't like to have the central TVMCI abstract class in LoopNode. It just does not belong there. It is going to become weirder and weirder if we add additional methods to it. LoopNode is an optional concept for the guest language. Since we cannot do the abstract CallTarget base class I am still opting for an abstract TruffleRuntime base class. Whats wrong with that?

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedOSRLoopNode.java
```diff
@@ -580,4 +589,45 @@ private void executeTransfer(FrameWithoutBoxing source, FrameWithoutBoxing targe
...
+                }
+            }
+        }
+
+        @SuppressWarnings("rawtypes")
+        Class<? extends TruffleLanguage> findLanguage(RootNode root) {
+            return super.findLanguageClass(root);
+        }
+
+        void initCallTarget(OptimizedCallTarget callTarget) {
```
chumer: Can we rename this to `onFirstExecution`? There is going to be a `onLoad `soon.

jtulach:
Accepting Christian's changes.

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTVMCI.java
```diff
@@ -0,0 +1,39 @@
...
+    @SuppressWarnings("rawtypes")
+    Class<? extends TruffleLanguage> findLanguage(RootNode root) {
+        return super.findLanguageClass(root);
+    }
+
+    void onFirstExecution(OptimizedCallTarget callTarget) {
+        super.onFirstExecution(callTarget.getRootNode());
+    }
+
+}
```
rschatz: Just noticed: This file has no newline at the end (eclipseformat should complain about that...)

jtulach: There was also missing license. Both problems fixed in most recent version.

jtulach:
I squashed all the changes into single commit.

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
```diff
@@ -305,7 +311,11 @@ public FrameInstance getCurrentFrame() {
         return iterateImpl(frame -> frame, 0);
     }
 
+    @Override
     public <T> T getCapability(Class<T> capability) {
+        if (capability.isAssignableFrom(TVMCI.class)) {
+            return capability.cast(tvmci);
+        }
```
woess: The `isAssignableFrom` check is backwards.

jtulach: Is it? I wrote a [test](https://github.com/jtulach/graal-core/commit/acfa1665abbea87c880bf71ae70d0df54d9b96a5) and it seems to work.

woess: erm, because it's testing the wrong thing. It shouldn't return tvmci for `Object.class`...

jtulach: Possibly. On the other hand the returned object can be casted without problems, so it is not completely bad. Do you want to modify the test to test what you think it should test?

woess: I've created #164 based on your test.

PR #136, opened 2016-03-16
Update .gitignore from .hgignore

tkrodriguez:
I pulled over a few rules from .hgignore to ignore more things by default.  Some of the things in .hgignore are no longer required because we split off jvmci.  Anything else showing up for anyone?

cc @axel22 @gilles-duboscq @rschatz @thomaswue


rschatz:
We could also add the `graal.options` file.

gilles-duboscq:
regarding
* `*.iml`
* `*.class`
* `*.jar`
Do we need them? I prefer to ignored the exact places where these things should be so that we notice when these files appear where they shouldn't

tkrodriguez:
We don't necessarily need them.  I view those as file kinds we'd never want to check into the repo so ignoring them everywhere is fine.  Would you like me to remove them?

tkrodriguez:
I added /mx.graal-core/graal.options as well.

tkrodriguez:
And *.bgv.

.gitignore
```diff
@@ -1,7 +1,30 @@
...
+*.cfg
+*.bgv
+*.jar
+*.zip
+*.rej
+*~
+core.*
 /mx.graal-core/env
 /mx.graal-core/eclipse-launches/
+/mx.graal-core/graal.options
```
zapster: I think this should be `/graal.options`. (cf #83)

gilles-duboscq:
> We don't necessarily need them. I view those as file kinds we'd never want to check into the repo so ignoring them everywhere is fine. Would you like me to remove them?
No strong opinion. Both have their virtues so i'm also fine with leaving them in.

.gitignore
tkrodriguez: Thanks.  I was guessing from reading the mx.py and thought it would be in the same location as the env.  Fixed.

PR #141, opened 2016-03-18
mx gate: make TestJMH task fail on error.

zapster:
Unfortunately, JMH always exits with zero return code per default even in case of an error. At least for JMH tasks in the gate we want fail-on-error behavior.


gilles-duboscq:
:+1:

PR #149, opened 2016-03-22
LSRA: skip block iteration when loop depth is 0.

zapster:


graal/com.oracle.graal.lir/src/com/oracle/graal/lir/alloc/lsra/LinearScanWalker.java
```diff
@@ -308,6 +308,10 @@ int findOptimalSplitPos(AbstractBlockBase<?> minBlock, AbstractBlockBase<?> maxB
 
         int minLoopDepth = maxBlock.getLoopDepth();
         for (int i = toBlockNr - 1; i >= fromBlockNr; i--) {
+            if (minLoopDepth == 0) {
+                // not possible to find a better position
+                break;
+            }
```
thomaswue: Small suggestion: move the check into the loop exit condition. minLoopDepth > 0 && i >= fromBlockNr

PR #150, opened 2016-03-23
Improve performance of node bit map and the block identification.

thomaswue:


graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/cfg/AbstractControlFlowGraph.java
```diff
@@ -22,11 +22,7 @@
  */
 package com.oracle.graal.compiler.common.cfg;
 
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
 
 public interface AbstractControlFlowGraph<T extends AbstractBlockBase<T>> {
```
dougxc: Do we still need `AbstractControlFlowGraph`? I see only one subclass. Although maybe there's an external client that extends it.

graal/com.oracle.graal.graph/src/com/oracle/graal/graph/NodeBitMap.java
```diff
@@ -32,13 +32,13 @@
 
     private long[] bits;
     private int nodeCount;
-    private final NodeIdAccessor nodeIdAccessor;
```
dougxc: Is there any check any more that a `NodeBitMap` (or anything else relying on stable node IDs) is not used across a graph compression?

dougxc:
I notice you've added `final` to a number of methods that have always been effectively final. Is this a suggestion we should adopt this more aggressive approach in general to save compilers from having to make speculations?

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
```diff
@@ -138,31 +178,12 @@ public Block blockFor(Node node) {
...
-    }
-
     public List<Loop<Block>> getLoops() {
         return loops;
     }
 
     private void identifyBlock(Block block) {
-        AbstractBeginNode start = block.getBeginNode();
-
-        // assign proxies of a loop exit to this block
```
dougxc: Did this code go somewhere else or is there no longer a requirement or good reason to assign proxies of a loop exit to the block?

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
```diff
@@ -181,133 +202,181 @@ private void identifyBlock(Block block) {
         }
     }
 
+    // Identify and connect blocks (including loop backward edges). Predecessors need to be in the
```
dougxc: Why not a javadoc style comment? Even though it's not public API, Eclipse shows javadoc in tooltips when hovering over a method call.

graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/cfg/AbstractControlFlowGraph.java
thomaswue: ControlFlowGraph includes the Graal graph specific parts. The backend only depends on AbstractControlFlowGraph to avoid the dependency on the Graal graph.

thomaswue:
I think it is generally advisable to put a final modifier if there is no clear use case for overwriting. It makes it clear that the method was written without being designed for overwriting. There is a small benefit in terms of simplifying optimizations, but this is a minor benefit.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
```diff
@@ -181,133 +202,181 @@ private void identifyBlock(Block block) {
...
+                for (int i = 1; i < predecessors.length; ++i) {
+                    probability += predecessors[i].probability;
+                }
+
+                if (block.getBeginNode() instanceof LoopBeginNode) {
+                    LoopBeginNode loopBegin = (LoopBeginNode) block.getBeginNode();
+                    probability *= loopBegin.loopFrequency();
                 }
             }
-            if (probability > 1. / MIN_PROBABILITY) {
```
dougxc: Why don't we need this any more?

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
thomaswue: The ControlFlowGraph generally only assigns the blocks for fixed nodes in contrast to the SchedulePhase, which also assigns blocks for all floating nodes. From a ProxyNode or PhiNode, one can always reach to the corresponding fixed node and ask for its block.

dougxc: Ok.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
thomaswue: Fixed.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/cfg/ControlFlowGraph.java
thomaswue: I will put it back in.

graal/com.oracle.graal.graph/src/com/oracle/graal/graph/NodeBitMap.java
thomaswue: No, this is at the moment not done.

dougxc: A quick scan of `NodeBitMap` uses seems to indicate that it could implement `AutoCloseable` and be used in try-with-resource statements where closing the map causes further operations to fail fast. In addition, if `NodeBitMap`s are only ever obtained from `Graph.createNodeBitMap()`, then one or two instances could be cached for reuse in a graph.

PR #152, opened 2016-03-23
RFC: more precise compile time benchmark state class

zapster:
`GraalCompilerState` is an extension of the `GraphState` class which also allows benchmarking of the different LIR stages. See included benchmarks for usage examples.

All suggestions are very welcome!


graal/com.oracle.graal.microbenchmarks/src/com/oracle/graal/microbenchmarks/lir/CompileTimeBenchmark.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.HashMap;
+
+import org.openjdk.jmh.annotations.Benchmark;
+
+import com.oracle.graal.code.CompilationResult;
+import com.oracle.graal.microbenchmarks.graal.GraalBenchmark;
+import com.oracle.graal.microbenchmarks.graal.util.MethodSpec;
+import com.oracle.graal.nodes.StructuredGraph;
+
+public class CompileTimeBenchmark extends GraalBenchmark {
```
zapster: IMHO benchmarks that call graal directly should be executed in hosted-mode to isolate harness and benchmark method. Maybe we should add a `@Fork(jvmArgsAppend = "-server")` here?

dougxc: How much do you expect timings to change depending on whether Graal is hosted or not? And does it matter since we care about both uses anyway.

tkrodriguez: I don't think hosted will really represent our compile times that well.  Hosted usually mean truffle and we know our compile times there are terrible but that's not primarily graal itself, just the expense of the huge graal graphs.  JVMCI builds are more likely to represent our actual compile times in a context where it matters.

graal/com.oracle.graal.microbenchmarks/src/com/oracle/graal/microbenchmarks/lir/GraalCompilerState.java
```diff
@@ -0,0 +1,444 @@
...
+
+    /**
+     * We only allow inner classes to subclass this to ensure that the {@link Setup} methods are
+     * executed in the right order.
+     */
+    @SuppressWarnings("try")
+    private GraalCompilerState() {
+        this.backend = Graal.getRequiredCapability(RuntimeProvider.class).getHostBackend();
+        this.providers = backend.getProviders();
+        this.suites = new DerivedOptionValue<>(this::createSuites);
```
tkrodriguez: You shouldn't have to do this any more.  The SuitesProvider is expected to do this for you now.

PR #155, opened 2016-03-23
Use VM independent word type checks when constructing `CallingConvention` of stubs.

rschatz:
That way, the VM specific word types (e.g. on hotspot `KlassPointer`, `MethodPointer` and `SymbolPointer`) can be used as arguments or return values of stubs or foreign calls.

This is currently unused. It will be used by the code for throwing `ClassCastException` from the compiler, which will be in my next PR. I'm putting up a separate PR for this to make it easier to review.

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotForeignCallLinkageImpl.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotForeignCallLinkageImpl.java
@@ -41,13 +41,14 @@ import jdk.vm.ci.meta.AllocatableValue;
 import jdk.vm.ci.meta.JavaType;
 import jdk.vm.ci.meta.LocationIdentity;
 import jdk.vm.ci.meta.MetaAccessProvider;
+import jdk.vm.ci.meta.ResolvedJavaType;
 import jdk.vm.ci.meta.Value;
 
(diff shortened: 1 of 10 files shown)


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/HotSpotForeignCallLinkageImpl.java
```diff
@@ -120,25 +122,25 @@ public static HotSpotForeignCallLinkage create(MetaAccessProvider metaAccess, Co
...
     }
 
-    private static JavaType asJavaType(Class<?> type, MetaAccessProvider metaAccess, CodeCacheProvider codeCache) {
-        if (WordBase.class.isAssignableFrom(type)) {
-            return metaAccess.lookupJavaType(codeCache.getTarget().wordJavaKind.toJavaClass());
-        } else {
-            return metaAccess.lookupJavaType(type);
+    private static JavaType asJavaType(Class<?> type, MetaAccessProvider metaAccess, WordTypes wordTypes) {
+        ResolvedJavaType javaType = metaAccess.lookupJavaType(type);
+        if (wordTypes.isWord(javaType)) {
```
rschatz: That's the actual change, everything else is just passing in the `WordTypes` object.

The `isWord` method may be overridden by VM specific code (e.g. on hotspot, not just `WordBase`, but also `KlassPointer`, `MethodPointer` and `SymbolPointer` count as "word" types).

PR #156, opened 2016-03-23
recompile stub for HotSpotNativeFunctionHandle if it was invalidated

dougxc:
This fixes test failures such as:
```
5) test1(com.oracle.nfi.test.NativeFunctionInterfaceTest)
jdk.vm.ci.common.JVMCIError: should not reach here: Execution of GNFI Callstub failed: os::dll_lookup
        at jdk.vm.ci.common.JVMCIError.shouldNotReachHere(JVMCIError.java:49)
        at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionHandle.call(HotSpotNativeFunctionHandle.java:95)
        at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionInterface.lookupFunctionPointer(HotSpotNativeFunctionInterface.java:117)
        at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionInterface.getFunctionHandle(HotSpotNativeFunctionInterface.java:81)
        at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionInterface.getFunctionHandle(HotSpotNativeFunctionInterface.java:103)
        at com.oracle.graal.truffle.hotspot.nfi.HotSpotNativeFunctionInterface.getFunctionHandle(HotSpotNativeFunctionInterface.java:39)
        at com.oracle.nfi.test.NativeFunctionInterfaceTest.test1(NativeFunctionInterfaceTest.java:115)
```

### DIFF
--- a/graal/com.oracle.graal.truffle.hotspot/src/com/oracle/graal/truffle/hotspot/nfi/HotSpotNativeFunctionHandle.java
+++ b/graal/com.oracle.graal.truffle.hotspot/src/com/oracle/graal/truffle/hotspot/nfi/HotSpotNativeFunctionHandle.java
@@ -26,7 +26,6 @@ import java.util.Arrays;
 
 import jdk.vm.ci.code.InstalledCode;
 import jdk.vm.ci.code.InvalidInstalledCodeException;
-import jdk.vm.ci.common.JVMCIError;
 import jdk.vm.ci.meta.JavaKind;
 
 import com.oracle.graal.debug.Debug;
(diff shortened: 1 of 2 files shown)


rschatz:
lgtm

graal/com.oracle.graal.truffle.hotspot/src/com/oracle/graal/truffle/hotspot/nfi/HotSpotNativeFunctionHandle.java
```diff
@@ -81,18 +82,24 @@ private void traceResult(Object result) {
     @Override
     public Object call(Object... args) {
         assert checkArgs(args);
-        try {
-            if (CompilerDirectives.inInterpreter()) {
-                traceCall(args);
+        while (true) {
```
gilles-duboscq: Should there be a limit to the number of times we go around this loop?
If this code is going away soon it might not be necessary, but if it's here to stay, maybe if it appears that we can never successfully install the stub it would be better to throw than to just loop forever.

tkrodriguez:
I believe hotspot has a way to keep certain nmethods from being swept because they are unused or from being invalidated by debug flags like DeoptimizeALot.  Maybe we could expose this capability through JVMCI?  I think it would require that the nmethod is compiled without dependencies since that defeats the whole purpose and no uncommon traps I think.

graal/com.oracle.graal.truffle.hotspot/src/com/oracle/graal/truffle/hotspot/nfi/HotSpotNativeFunctionHandle.java
dougxc: I thought about that but we should soon have the VM independent mechanism for this so this is a "hot patch". That said, I'll put a counter in to fail after 10 attempts just to avoid the possibility of an endless loop.

dougxc:
If possible, we should not interfere with how the VM wants to manage the code cache. I assume it takes the hotness of a method into account in which case we're ok.

tkrodriguez:
Well it's used for things like the linkTo* methods that lambda uses and are kept around for similar reasons.  But it sounds like it's not really necessary here.

PR #157, opened 2016-03-23
TraceRA: combine SSI construction and interval building.

zapster:
This PR contains a new experimental phase which does SSI construction and interval building in a single phase.
pro side: we avoid a separate lifetime analysis phase i.e. one traversal over the operands which is quite expensive.
con side: we store the interval data for all traces simultaneously, which increases the data that is alive.


rschatz:
new commits also lgtm

zapster:
thanks for the review @rschatz

zapster:
I reordered the last 4 commits to ease potential bisecting. Overall diff is the same.

PR #159, opened 2016-03-23
Debug.ENABLED should only be true if scoped debugging is in use

dougxc:
When removing use of the `jvmci.debug.enable` system property for initializing `Debug.ENABLED`, other options were [consulted](https://github.com/graalvm/graal-core/commit/57c9f29606b26d76a6817abaa5a7157409ca8131#diff-44907a377a861a8893470d53ce06270dR71) that shouldn't have been.

This addresses the regression in certain Graal compilation time benchmarks since the above commit was made on Jan 11.


tkrodriguez:
looks good

PR #161, opened 2016-03-24
support for throwing ClassCastException and ArrayStoreException from compiled code

rschatz:
Depends on https://bugs.openjdk.java.net/browse/JDK-8152634

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/meta/DefaultHotSpotLoweringProvider.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/meta/DefaultHotSpotLoweringProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2016, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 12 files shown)


graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/meta/DefaultHotSpotLoweringProvider.java
```diff
@@ -542,37 +543,51 @@ private void lowerOSRStartNode(OSRStartNode osrStart) {
...
 
-    private void lowerBytecodeExceptionNode(BytecodeExceptionNode node) {
+    private boolean throwCachedException(BytecodeExceptionNode node) {
+        Throwable exception;
+        if (node.getExceptionClass() == NullPointerException.class) {
+            exception = Exceptions.cachedNullPointerException;
+        } else if (node.getExceptionClass() == ArrayIndexOutOfBoundsException.class) {
+            exception = Exceptions.cachedArrayIndexOutOfBoundsException;
+        } else {
+            return false;
```
rschatz: It doesn't really make sense to support cached class cast or array store exceptions, since the message would be missing.

That's by the way also true for the `ArrayIndexOutOfBoundsException`. Is that a bug?

dougxc: if the `OmitHotExceptionStacktrace` option has been enabled, then the user has to be ok with "degraded" exceptions.

graal/com.oracle.graal.replacements.test/src/com/oracle/graal/replacements/test/ArrayStoreBytecodeExceptionTest.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }, Exceptions.class, "throwArrayStore", Object.class);
+    }
+
+    public static void arrayStoreSnippet(Object obj) {
+        Exceptions.throwArrayStore(obj);
+    }
+
+    @Parameter(0) public Object object;
+    @Parameter(1) public Class<?> cls;
```
dougxc: The `cos` field appears to be unused.

rschatz: The field is indeed unused, but the parameter is used in the `name` property, to give a nice output when running `mx unittest --verbose`. The type of the object is a lot more useful than the instance for this test.

The field is needed because otherwise JUnit complains about an unused parameter.

zapster: You could also replace the `@Parameter` fields with the constructor variant:
```
diff --git a/graal/com.oracle.graal.replacements.test/src/com/oracle/graal/replacements/test/ArrayStoreBytecodeExceptionTest.java b/graal/com.oracle.graal.replacements.test/src/com/oracle/graal/replacements/test/ArrayStoreBytecodeExceptionTest.java
index 9051a3b..004a115 100644
--- a/graal/com.oracle.graal.replacements.test/src/com/oracle/graal/replacements/test/ArrayStoreBytecodeExceptionTest.java
+++ b/graal/com.oracle.graal.replacements.test/src/com/oracle/graal/replacements/test/ArrayStoreBytecodeExceptionTest.java
@@ -63,8 +63,12 @@ public class ArrayStoreBytecodeExceptionTest extends BytecodeExceptionTest {
         Exceptions.throwArrayStore(obj);
     }
 
-    @Parameter(0) public Object object;
-    @Parameter(1) public Class<?> cls;
+    ArrayStoreBytecodeExceptionTest(Object object, Class<?> cls) {
+        Assert.assertEquals("Class must match!", object.getClass(), cls);
+        this.object = object;
+    }
+
+    private Object object;
 
     @Parameters(name = "{1}")
     public static Collection<Object[]> data() {

```

But basically it only replaces an unused field with an unused parameter in the constructor.

PR #162, opened 2016-03-24
re-added ControlFlowGraph.frequencyFor

dougxc:


dougxc:
This change is unnecessary.

PR #163, opened 2016-03-24
Make OptimizedCallTarget#log static.

zapster:
This fixes an eclipse 4.5.2 warning which was introduce by #135.


graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
```diff
@@ -135,7 +135,7 @@ private static GraalTruffleRuntime runtime() {
         return (GraalTruffleRuntime) Truffle.getRuntime();
     }
 
-    public final void log(String message) {
+    public static final void log(String message) {
```
woess: shouldn't checkstyle complain about a redundant `final` here?

zapster: Also though so but is seems that `static final` is not redundant as it prohibits method hiding.

PR #164, opened 2016-03-24
fix check against TVMCI.class in getCapability

woess:


graal/com.oracle.graal.truffle.test/src/com/oracle/graal/truffle/test/TruffleRuntimeTest.java
```diff
@@ -62,6 +62,11 @@ public void findTVMCI() {
...
 
         Object object = runtime.getCapability(Object.class);
-        assertSame("The same instance returned for Object.class", tvmci, object);
+        assertNull("null returned for Object.class", object);
+
+        abstract class TVMCISubclass extends TVMCI {
+        }
+        TVMCISubclass subclass = runtime.getCapability(TVMCISubclass.class);
+        assertNull("null returned for TVMCI subclass", subclass);
     }
```
jtulach: Good test. I think it should be part of the test suite: https://github.com/jtulach/graal-core/commit/08c54ff1b11fbf073e00196d4ae1a6d9fe53916b

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/GraalTruffleRuntime.java
```diff
@@ -313,7 +313,7 @@ public FrameInstance getCurrentFrame() {
 
     @Override
     public <T> T getCapability(Class<T> capability) {
-        if (capability.isAssignableFrom(TVMCI.class)) {
+        if (capability == TVMCI.class) {
             return capability.cast(tvmci);
```
jtulach: However as https://github.com/jtulach/graal-core/commit/08c54ff1b11fbf073e00196d4ae1a6d9fe53916b shows, you don't need this fix to make your test run.

jtulach:
What is the goal of this change?

As far as I remember it started with a claim that the `isAssignableFrom` in `getCapability` is up-side-down. So far we haven't found a test that would prove that!

What are your next plans?

woess:
Did you see my following change in the test?
```diff
-        assertSame("The same instance returned for Object.class", tvmci, object);
+        assertNull("null returned for Object.class", object);
```
Yes, my bad, it wasn't upside down. But imho it's wrong to return `tvmci` if you pass in `Object.class`. I thought a `==` comparison is better here anyway, so I changed it to that. Feel free to suggest a different fix (would you rather check for this case explicitly?).

jtulach:
Yes, I can agree with that. Once there is more than one capability, querying for `Object.class` makes little sense. Using `==` check to make sure `getCapability(Object.class)` returns `null` is OK.

PR #165, opened 2016-03-25
Improve AddNode/MulNode input swapping.

christhalinger:
While looking at the micro-benchmarks of https://bugs.openjdk.java.net/browse/JDK-8074124 I noticed that Graal generated a useless move in a very tight loop:

```
  0x000000011c8a3fd9: movzwl 0x10(%r10,%r8,2),%ecx
  0x000000011c8a3fdf: add    %r9d,%ecx
  0x000000011c8a3fe2: inc    %r8d
  0x000000011c8a3fe5: mov    %ecx,%r9d          ;*iload_2 {reexecute=0 rethrow=0 return_oop=0}
                                                ; - org.openjdk.UnsafeConvBench::plain@4 (line 210)

  0x000000011c8a3fe8: cmp    %r8d,%eax
  0x000000011c8a3feb: jg     0x000000011c8a3fd0  ;*ireturn {reexecute=0 rethrow=0 return_oop=0}
                                                ; - org.openjdk.UnsafeConvBench::plain@28 (line 213)
```

It turned out the logic in `AddNode.generate` swapped the inputs solely based on the property if `y` was dead.  In this case both inputs are dead but one of them is an induction variable phi.  This patch takes that into consideration.  Now the loop looks like this:

```
  0x000000010e3bb9d9: movzwl 0x10(%r10,%r8,2),%ecx
  0x000000010e3bb9df: add    %ecx,%r9d
  0x000000010e3bb9e2: inc    %r8d               ;*iload_2 {reexecute=0 rethrow=0 return_oop=0}
                                                ; - org.openjdk.UnsafeConvBench::plain@4 (line 210)

  0x000000010e3bb9e5: cmp    %r8d,%eax
  0x000000010e3bb9e8: jg     0x000000010e3bb9d0  ;*ireturn {reexecute=0 rethrow=0 return_oop=0}
                                                ; - org.openjdk.UnsafeConvBench::plain@28 (line 213)
```

### DIFF
--- a/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/calc/BinaryArithmeticNode.java
+++ b/graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/calc/BinaryArithmeticNode.java
@@ -25,9 +25,6 @@ package com.oracle.graal.nodes.calc;
 import java.io.Serializable;
 import java.util.function.Function;
 
-import jdk.vm.ci.common.JVMCIError;
-import jdk.vm.ci.meta.Constant;
-
 import com.oracle.graal.compiler.common.type.ArithmeticOpTable;
 import com.oracle.graal.compiler.common.type.ArithmeticOpTable.BinaryOp;
 import com.oracle.graal.compiler.common.type.Stamp;
@@ -42,8 +39,12 @@ import com.oracle.graal.nodes.ArithmeticOperation;
 import com.oracle.graal.nodes.ConstantNode;
 import com.oracle.graal.nodes.StructuredGraph;
 import com.oracle.graal.nodes.ValueNode;
+import com.oracle.graal.nodes.ValuePhiNode;
 import com.oracle.graal.nodes.spi.ArithmeticLIRLowerable;
 import com.oracle.graal.nodes.spi.NodeValueMap;
+
+import jdk.vm.ci.common.JVMCIError;
+import jdk.vm.ci.meta.Constant;
 
 @NodeInfo
 public abstract class BinaryArithmeticNode<OP> extends BinaryNode implements ArithmeticOperation, ArithmeticLIRLowerable, Canonicalizable.Binary<ValueNode> {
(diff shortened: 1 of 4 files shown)


christhalinger:
Should the same logic be done on all `BinaryCommutative` nodes?

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/calc/BinaryArithmeticNode.java
```diff
@@ -292,4 +284,28 @@ public BinaryNode maybeCommuteInputs() {
...
+    /**
+     * Determines if it would be better to swap the inputs in order to produce better assembly code.
+     * First we try to pick a value which is dead after this use. If both values are dead at this
+     * use then we try pick an induction variable phi to encourage the phi to live in a single
+     * register.
+     *
+     * @param nodeValueMap
+     * @return true if inputs should be swapped, false otherwise
+     */
+    protected boolean shouldSwapInputs(NodeValueMap nodeValueMap) {
```
christhalinger: I think this method should rather be in the `BinaryCommutative` interface.

thomaswue: Yes, this would be better as it is only relevant for commutative operations.

thomaswue:
We should probably add for what type of assembly instruction sets the swapping is beneficial - i.e., the ones where the result and one of the inputs need to have the same machine location.

thomaswue:
In general, looks good. Ideal would be if we could also have a unit test as a regression here is hard to detect.

tkrodriguez:
I'm not sure we can build a regression test for this.  I think the extra move may be generated in the assembly for CommutativeOp, though I don't know if it's the case here.

<pre>
        @Override
        public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
            AllocatableValue input;
            if (sameRegister(result, y)) {
                input = x;
            } else {
                AMD64Move.move(crb, masm, result, x);
                input = y;
            }
</pre>

I guess we could directly examine the LIR.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/calc/BinaryArithmeticNode.java
christhalinger: Good we have default methods :-)

christhalinger:
Do you have existing tests that do that?

tkrodriguez:
I don't know of any examples.

tkrodriguez:
If it looks hard build a test then don't worry about it.

PR #172, opened 2016-03-28
Rationalize dump levels and change default to 1

tkrodriguez:
I made the levels more explicit and forced the call sites to actively pick one.  Most of the ones using dump stayed at level 2 but I tried to change all debugging and testing ones to 1.  I also added support for symbolic names in the dump specifiers so you can say -G:Dump=:verbose to get level 3.  I also adjusted places where a constant integer was being passed instead of a symbolic name.

cc @dougxc @gilles-duboscq @zapster @thomaswue

### DIFF
--- a/graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/alloc/TraceStatisticsPrinter.java
+++ b/graal/com.oracle.graal.compiler.common/src/com/oracle/graal/compiler/common/alloc/TraceStatisticsPrinter.java
@@ -31,12 +31,11 @@ import com.oracle.graal.debug.Indent;
 
 public final class TraceStatisticsPrinter {
     private static final String SEP = ";";
-    private static final int TRACE_DUMP_LEVEL = 3;
 
     @SuppressWarnings("try")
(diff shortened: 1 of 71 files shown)


zapster:
I like the change, just a few remarks about the naming. For me it is not really intuitive that that `info` is more verbose than `basic`. Also, for new users of the API it might not be clear that `basic` is the default and that it is supposed to be the least verbose level. Maybe it just me being spoiled by other systems using similar names.

dougxc:
lgtm

tkrodriguez:
I wasn't totally happy with the difference between the names either.  The intent was that basic was the minimum, info was extra stuff, and verbose is stuff you rarely care about.  I'm up for a renaming if you have better suggestions.

dougxc:
The naming doesn't really bother me but `TERSE_LOG_LEVEL` is one suggestion.

tkrodriguez:
I don't feel like that's much of an improvement.  Adopting something stolen from log4j would give us INFO,DEBUG,TRACE with increasing levels of verbosity.   I don't like that much either.  Anyway, if someone wants something different feel free to propose it but I'll leave it alone for now.

PR #174, opened 2016-03-29
Ensure that Graphs end up in a unique scope when dumping

tkrodriguez:
This ensures that folders in an IGV dump are unique per graph based on the context of the first time a graph is seen.  It has the side effect that if you dump A, dump B and then dump A again you'll get 3 folders, so the first and last folder will actually correspond to the same graph but at least you'll know that within a folder it's all the same graph.

@rschatz


rschatz:
Seems this somehow messed up the dumping of stubs. They are now dumped into a folder called `main: Top Scope`, with no mention of the stub name.

tkrodriguez:
hmm.  I'll take a look.

tkrodriguez:
It appears we've always dumped the first graph of stubs into main: Top Scope but all the rest ended up in the right place so we never noticed I guess.  Now that we stick with the first name we get it wrong.

PR #178, opened 2016-03-30
Introduce `DynamicPiNode` to capture improved stamp when a dynamic ca…

rschatz:
…st becomes static later in the compilation.

In contrast to the normal `PiNode`, the `DynamicPiNode` will stick around, even though it doesn't actually improve the stamp. Otherwise we lose the opportunity to improve the stamp later when the input becomes constant.

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/meta/HotSpotGraphBuilderPlugins.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/meta/HotSpotGraphBuilderPlugins.java
@@ -48,7 +48,6 @@ import sun.reflect.Reflection;
 
 import com.oracle.graal.api.replacements.SnippetReflectionProvider;
 import com.oracle.graal.compiler.common.spi.ForeignCallsProvider;
-import com.oracle.graal.compiler.common.type.StampFactory;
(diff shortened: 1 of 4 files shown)


lukasstadler:
looks good, thanks for the fix.

tkrodriguez:
I'd talked a bit with thomas when he was doing these changes and we discussed the idea that PiNodes should always know the condition that guards them and the stamp they produce should be driven from that test.  Right now PiNodes are sort of unsafe since the condition and the resulting type are unrelated.  That seems very much like the same kind of problem we're having here.

lukasstadler:
the solution is similar, just that the issue that this change solves is about performance, and not safety: you might still improve the stamp as long as the class is not constant, and you can only do that if you maintain the connection.

tkrodriguez:
That wasn't a complaint about this change.  It's good and reasonable but it points at a general problem with PiNodes being disconnected from the test that proves them.

rschatz:
But they are still connected to the guard, right?

tkrodriguez:
Yes.  The idea would be take advantage of the connection to the guard to use the LogicNode getSucceedingStamp* helpers to drive the stamp of the PiNode.  So for this example the InstanceOfDynamic becomes InstanceOf and getSucceedingStamp() returns the proper stamp that we need.  There are probably some difficult details buried down in there but there's something appealing about the safety and power of this approach.  It also doesn't cover all Pi uses either.

PR #179, opened 2016-03-30
OptimizedCallTarget fixes

eregon:
Essentially simplification of the code and more explicit assumptions.


graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/OptimizedCallTarget.java
```diff
@@ -208,8 +208,11 @@ private void initialize() {
     }
 
     private void ensureCloned() {
-        if (uninitializedRootNode == UNINITIALIZED) {
-            this.uninitializedRootNode = sourceCallTarget == null ? cloneRootNode(rootNode) : sourceCallTarget.uninitializedRootNode;
+        if (sourceCallTarget != null) {
```
smarr: Does this method deserve a comment that it should be used while holding the lock? It is currently only used once, but, I am very much in favor of documenting locking policies.

PR #184, opened 2016-05-18
Update to emit assembler through common interface and addition of AVX…

mcberg2016:
… emit support including changes for ArrayEquals.

### DIFF
--- a/graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
+++ b/graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
@@ -167,7 +167,6 @@ public class AMD64Assembler extends Assembler {
      * Constants for X86 prefix bytes.
      */
     private static class Prefix {
-
         private static final int REX = 0x40;
         private static final int REXB = 0x41;
         private static final int REXX = 0x42;
@@ -184,6 +183,42 @@ public class AMD64Assembler extends Assembler {
         private static final int REXWRB = 0x4D;
         private static final int REXWRX = 0x4E;
         private static final int REXWRXB = 0x4F;
+        private static final int VEX_3BYTES = 0xC4;
+        private static final int VEX_2BYTES = 0xC5;
+    }
+
+    private static class VexPrefix {
+        private static final int VEX_R = 0x80;
+        private static final int VEX_W = 0x80;
+    }
+
+    private static class AvxVectorLen {
+        private static final int AVX_128bit = 0x0;
+        private static final int AVX_256bit = 0x1;
+    }
+
+    private static class VexSimdPrefix {
+        private static final int VEX_SIMD_NONE = 0x0;
+        private static final int VEX_SIMD_66 = 0x1;
+        private static final int VEX_SIMD_F3 = 0x2;
+        private static final int VEX_SIMD_F2 = 0x3;
+    }
+
+    private static class VexOpcode {
+        private static final int VEX_OPCODE_NONE = 0x0;
+        private static final int VEX_OPCODE_0F = 0x1;
+        private static final int VEX_OPCODE_0F_38 = 0x2;
+        private static final int VEX_OPCODE_0F_3A = 0x3;
+    }
+
+    private AMD64InstructionAttr curAttributes;
+
+    AMD64InstructionAttr getCurAttributes() {
+        return curAttributes;
+    }
+
+    void setCurAttributes(AMD64InstructionAttr attributes) {
+        curAttributes = attributes;
     }
 
     /**
@@ -498,6 +533,7 @@ public class AMD64Assembler extends Assembler {
                 emitInt(disp);
             }
         }
+        setCurAttributes(null);
     }
 
     /**
@@ -511,9 +547,9 @@ public class AMD64Assembler extends Assembler {
 
         private final String opcode;
 
-        private final int prefix1;
-        private final int prefix2;
-        private final int op;
+        protected final int prefix1;
+        protected final int prefix2;
+        protected final int op;
 
         private final boolean dstIsByte;
         private final boolean srcIsByte;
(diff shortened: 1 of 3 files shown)


graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -731,14 +1072,122 @@ protected AMD64MROp(String opcode, int prefix1, int prefix2, int op, OpAssertion
         @Override
         public final void emit(AMD64Assembler asm, OperandSize size, Register dst, Register src) {
             assert verify(asm, size, src, dst);
-            emitOpcode(asm, size, getRXB(src, dst), src.encoding, dst.encoding);
-            asm.emitModRM(src, dst);
+            boolean isSimd = false;
+            boolean noNds = false;
+
+            // TODO: add opcodes from the initializer here
+            switch (op) {
```
dougxc: I will review this in detail later.

rschatz:
Thanks for this pull request!

After an inital look, I have a few general comments:

It seems in the new versions of the `AMD64*Op.emit` methods, you have logic to automatically use AVX instructions in place of SSE instructions, but I don't see a CPU feature flag test. Does that work on machines that don't have AVX?

On a more general note, this automatic selection logic is missing an opportunity. We're using these instructions as if they were two-operand instructions, i.e. we force the result and the first input to be in the same register, possibly inserting an extra `MOV` instruction. If we move the selection logic a bit higher, to the LIR code, we have the opportunity to allocate the result and the first input to different registers.

I'll do a more detailed review on monday...

mcberg2016:
Roland,

The logic you are looking for is in InstructionAttr in the default constructor, where if we run on a machine without AVX you will get standard SSE code, of course at some point a control like UseAVX as a user flag like in c1/c2 will also be a nice to have to give users the ability to choose assembler emit.

I think for the selection logic another class could be introduced, an AVXOp class, which could have its own emit which could handle the code as a both 2 and 3 address instruction emit on AVX and greater enabled targets, as both forms persist.   I could tackle that in a later revision.  Mostly this code is being added to get the ball rolling so to speak.

Regards,
Michael

graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -184,6 +183,202 @@ public String toString() {
...
+            if (supports(CPUFeature.AVX2)) {
+                useAvx = 2;
+            } else if (supports(CPUFeature.AVX)) {
+                useAvx = 1;
+            } else {
+                useAvx = 0;
+            }
+        }
+
+        public void initAttr(
```
rschatz: Now I see why I missed the AVX feature check at first. This method is used to communicate into the `simdPrefix` method via global variables. This style of programming is very hard to read, I would prefer if these things were made explicit arguments.

Also, many of these flags look unused.

graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -511,9 +706,9 @@ protected void emitOperandHelper(int reg, AMD64Address addr, boolean force4Byte)
 
         private final String opcode;
 
-        private final int prefix1;
-        private final int prefix2;
-        private final int op;
+        public final int prefix1;
+        public final int prefix2;
+        public final int op;
```
rschatz: I think these can be made protected.

graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -1894,6 +2321,126 @@ private int prefixqAndEncode(int regEncoding, int rmEncoding) {
         return regEnc << 3 | rmEnc;
     }
 
+    private void vexPrefix(boolean vexR, boolean vexB, boolean vexX, int ndsEncoding, int pre, int opc) {
```
rschatz: If you use a single argument `int rxb` instead of three booleans, you can re-use the existing functions `getRXB(Register, Register)` and `getRXB(Register, AMD64Address)`...

rschatz:
The user selection logic is already implicitly in there, if you run the VM with `-XX:UseAVX=0`, then `CPUFeature` set will not report AVX support.

Regarding the separate `AVXOp` for 3-operand encodings, I agree we should do this in a separate change.

See also inline comments in the patch.

Other than that, looks good.

graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
mcberg2016: This is a bit of an interesting problem, even in C++ i had to make a reference to the current attributes object when redesigning the x86 assembler, which had to remain live until the end of scope of the current instruction.  Rationale: EVEX address modes will be calculated after the simd and vex prefix calls where attribute info must be conveyed into emitOperandHelper.  I went this route here to avoid making as many InstructionAttr objects as there were instructions minted, as it would complicate garbage collection.  It seemed to me I could save a good deal of process time and memory by just retargeting the fields for each use on an instruction.  Since each masm/asm object is thread local, it is safe as the instructionAttr object is local in instance to it.

mcberg2016:
I'll re-add my comment regarding the InstructionAttr object.  In C2, when I redisgned the x86 assembler I managed this object as local scope and destructed at the instruction side.  I looked at using a new object for every instruction here, but thought it would complicate garbage collection and use Nxsizeof(InstructionAttr) size where N is the number of simd instuctions, and so would up using a thread safe instance within the assembler object which now only modified attributes for simd instructions.  It gets further complicated by the notion that for address based simd instructions, the emit operand helper routine will need the instruction attribute info on EVEX to accomplish compressed displacement offset calculations which are part of the machines address model.  When I add the EVEX layer all of the unused fields will be utilized, along with a considerable amount of additional support.  I am just preparing for that eventuality.

dougxc:
I don't see how `AMD64Assembler.attributes` is thread safe since it's a plain value as opposed to a `ThreadLocal`. It will be shared across all compiler threads.

graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -184,9 +183,193 @@ public String toString() {
         private static final int REXWRB = 0x4D;
         private static final int REXWRX = 0x4E;
         private static final int REXWRXB = 0x4F;
+        private static final int VEX_3BYTES = 0xC4;
+        private static final int VEX_2BYTES = 0xC5;
+    }
+
+    private static class VexPrefix {
+        private static final int VEX_B = 0x20;
+        private static final int VEX_X = 0x40;
```
rschatz: `VEX_B` and `VEX_X` are unused. Did you edit this in eclipse? At least eclipse is giving a warning on that, I don't know about javac. We enforce zero warnings in our gate.

If you don't use eclipse, you can test this yourself with `mx build --warning-as-error --jdt ~/ecj/ecj-4.5.1.jar`.

rschatz:
The whole `AMD64Assembler` object is local to the compilation, so at least regarding thread safety we're ok.

I still think the "global" object makes the code hard to read, and we should do this as an argument. For example, there is already the `pre` argument for the `pp` field of the prefix. Why not just add arguments for the `W` and `L` fields, too? As for the other (currently unused) fields, I think that's something that could be placed in the `AVXOp` class, once we have it.

On a more general note: We try to have descriptive messages in the first line of the commit message (e.g. "Add AVX support to AMD64Assembler."), because that's what you see in most git UIs. Can you please use that commit message?

mcberg2016:
There is an intermediate path to take, passing attribute as an argument for each instruction. It will then be allocated it so that its local to each and every instructions scope.  I did this in C1/C2.  It was nice there as the object was destroyed for each scope exit of an instruction.  Since there are two paths to get into the assembler, one via the AMD64Op hierarchy and the other directly as called through the assembler object, its seems best to keep the attributes instruction local.  Also, there are more than just the 5 fields to add as you will see in EVEX, so the call signatures will be much larger if not object encapsulated.

Ok, I will use that in the commit message, along with some addendum for each update.

PR #185, opened 2016-06-01
Test to avoid evil in Graal

ericsedlar:


woess:
There's no evil in Graal.

PR #187, opened 2016-06-08
Addition of libm log and log2 plus assembler support

mcberg2016:


graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64MathIntrinsicOp.java
```diff
@@ -44,23 +59,736 @@
...
+        masm.pand(temp5, temp1);
+        masm.pand(dest, temp6);
+        masm.subsd(temp1, temp5);
+        masm.mulpd(temp5, dest);
+        masm.andl(gpr1, 32752);
+        masm.subl(gpr1, gpr2);
+        masm.cvtsi2sdl(temp7, gpr1);
+        masm.mulsd(temp1, dest);
+        crb.compilationResult.recordDataPatch(masm.position(), logTwoDataRef);
+        masm.movdq(temp6, masm.getPlaceholder());     // 0xfefa3800, 0x3fa62e42
```
tkrodriguez: This idiom should probably be handled by creating a subclass of DataPointerConstant that wraps the array and does any endianness conversion.

graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64MathIntrinsicOp.java
```diff
@@ -44,23 +59,736 @@
...
     }
 
     @Override
     public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
-        switch (opcode) {
-            case LOG:   masm.flog(asRegister(result, AMD64Kind.DOUBLE), asRegister(input, AMD64Kind.DOUBLE), false); break;
-            case LOG10: masm.flog(asRegister(result, AMD64Kind.DOUBLE), asRegister(input, AMD64Kind.DOUBLE), true); break;
+        switch(opcode) {
+            case LOG:   logIntrinsic(asRegister(result, AMD64Kind.DOUBLE), asRegister(input, AMD64Kind.DOUBLE), crb, masm); break;
+            case LOG10: log10Intrinsic(asRegister(result, AMD64Kind.DOUBLE), asRegister(input, AMD64Kind.DOUBLE), crb, masm); break;
```
tkrodriguez: I don't think we really want to emit all that assembly every time it's called.  Instead it should be more like what hotspot does where it's generated into a stub and called as a function.  I don't know if we have any examples of platform independent stubs that you could copy though.  For now it will probably need to be a hotspot SnippetStub and we'll need a new Node and Op to drive the code generation in the snippet.  Roland or Doug is there a simpler way that I'm not seeing?

graal/com.oracle.graal.lir/src/com/oracle/graal/lir/asm/ArrayDataPointerConstant.java
```diff
@@ -0,0 +1,78 @@
...
+        this.byteBuffer = ByteBuffer.allocate(array.length * 4);
+        serialize(byteBuffer);
+    }
+
+    @Override
+    public boolean isDefaultForKind() {
+        return false;
+    }
+    
+    public void serialize(ByteBuffer buffer) {
```
mcberg2016: I will likely be adding an enum to this class to process various array types with under a switch for serialize and getSerializedSize.

mcberg2016:
Tom is now a collaborator on this branch and change set.  He is adding the stub support for me to plug into.

dougxc:
Do we already have unit tests for all the interesting input values of these libm operations? If not, they should either be added in this PR (preferably) or in a follow PR.

graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64LibmOp.java
```diff
@@ -0,0 +1,785 @@
...
+
+import jdk.vm.ci.amd64.AMD64;
+import jdk.vm.ci.amd64.AMD64Kind;
+import jdk.vm.ci.amd64.AMD64.CPUFeature;
+import jdk.vm.ci.meta.AllocatableValue;
+import jdk.vm.ci.meta.Value;
+import jdk.vm.ci.code.Register;
+import jdk.vm.ci.code.site.DataSectionReference;
+
+// @formatter:off
```
christhalinger: Please turn off the formatter only for the sections you want, not the whole class.

graal/com.oracle.graal.lir/src/com/oracle/graal/lir/asm/ArrayDataPointerConstant.java
```diff
@@ -0,0 +1,78 @@
...
+import jdk.vm.ci.meta.SerializableConstant;
+import jdk.vm.ci.code.site.DataSectionReference;
+
+/**
+ * Base class for {@link Constant constants} that represent a pointer to the data section.
+ */
+public class ArrayDataPointerConstant extends DataPointerConstant {
+
+    public DataSectionReference dataRef;
+    public CompilationResultBuilder crb;
```
christhalinger: I don't think these should be public.

tkrodriguez:
There are Math_* tests that test all the special values.  There are good tests for these functions in the JDK tests in test/java/lang/Math like http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/tip/test/java/lang/Math/Log1pTests.java.  We could consider modifying them as unit tests.

PR #189, opened 2016-06-23
Update directives instrumentation API

mur47x111:
Update directives instrumentation API as well as the underlying mechanism. 
- Move node classes to the package com.oracle.graal.nodes.debug.instrumentation
- Create mapping from BCI to nodes in BytecodeParser. This is for precisely locating the instrumentation target.
- Thanks to the precise bytecode mapping, InstrumentationNode and instrumentation-related phases are refined. Also, the redundant directive API method for inspecting invocation and the additional boolean parameter to the constructor of InstrumentationBeginNode are not needed.

Add comments to related classes. 
Add multiple tests in com.oracle.graal.api.directives.test


tkrodriguez:
So the process is to review this in the open but then push it through ol-bitbucket and close it on github?

tom

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
```diff
@@ -367,65 +367,58 @@ public static void ensureVirtualizedHere(@SuppressWarnings("unused") Object obje
...
      *
      * </blockquote>
      *
-     * @param offset the length of a sequential path from a target node to the invocation to this
-     *            API (if negative), or from the invocation to {@link #instrumentationEnd()} to a
-     *            target node (if positive). Pass 0 to anchor the instrumentation.
+     * @param offset the length of a sequential path from a target bytecode to the invocation to
+     *            this API (if negative, excluding the bytecode loading this parameter), or from the
+     *            invocation to {@link #instrumentationEnd()} to a target node (if positive). Pass 0
+     *            to anchor the instrumentation.
```
dougxc: The term "node" makes no sense when talking about bytecode instructions. And is "target node" the same as "target bytecode" here? If so, then all uses should be replaced with the "instrumented bytecode instruction" or  the "bytecode instruction being instrumented". Also, why `excluding the bytecode loading this parameter`? In general, I think the API of this method is confusing. I think you should introduce an `InstrumentationTarget` enum that denotes what can be targeted:

```
enum InstrumentationTarget { NONE, NEW, MONITORENTER, MONITOREXIT }
```
and recast this method as follows:
```
    public static void instrumentationBegin(InstrumentationTarget target, int precedingOccurrence)
```

That is, instrumentation must always come after the targeted instruction (like it does in all your tests). The special value `NONE` means no instruction is targeted. Otherwise `precedingOccurrence ` denotes the N'th preceeding instance of the instruction matched by `target` (0 is first preceding instance, 1 is second... etc)

dougxc:
Could you please expand the PR description to explain the major changes in more detail. A PR this size is hard to navigate otherwise.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
```diff
@@ -367,65 +367,58 @@ public static void ensureVirtualizedHere(@SuppressWarnings("unused") Object obje
...
+     * @see #instrumentationBegin(int)
      */
     public static void instrumentationEnd() {
     }
 
     /**
-     * @return an integer representing a runtime path taken for a @Snippet. This API is valid only
-     *         if invoked within an instrumentation (see {@link #instrumentationBegin(int)} and
+     * @return an integer representing a control flow path taken for a @Snippet. This API is valid
+     *         only if invoked within an instrumentation (see {@link #instrumentationBegin(int)} and
```
dougxc: don't use the word API here

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
```diff
@@ -367,65 +367,58 @@ public static void ensureVirtualizedHere(@SuppressWarnings("unused") Object obje
...
      *         {@link #instrumentationEnd()} , and the associated target node of the instrumentation
-     *         is a preceding node that will be substituted by a @Snippet with multiple runtime
+     *         is a preceding node that will be substituted by a @Snippet with multiple control flow
      *         paths. It will be replaced with a ValuePhiNode with constant integer inputs [0, N-1],
-     *         where N denotes the number of runtime paths of the snippet.
+     *         where N denotes the number of control flow paths of the snippet.
```
dougxc: Don't mention ValuePhiNode - that's an implementation detail that may change and adds no value here.
Neither the javadoc nor the tests exercising this method make it clear how this value can be useful.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
tkrodriguez: I agree.  You really can't use bytecode offsets at the source level with any confidence.  Even specifying precedingOccurrence seems dubious.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
```diff
@@ -71,19 +81,24 @@ static long arrayBaseOffset() {
...
+        Word cArray = CStringConstant.cstring(value);
         while (i-- > 0) {
             // assuming it is ASCII string
             // array[i] = (char) cArray.readByte(i);
             UNSAFE.putChar(array, arrayBaseOffset() + i * 2, (char) cArray.readByte(i));
         }
         String newString = (String) newInstance(String.class, false);
         UNSAFE.putObject(newString, valueOffset(), array);
-        UNSAFE.putInt(newString, hashOffset(), 0);
+        UNSAFE.putInt(newString, hashOffset(), hashOf(value));
```
tkrodriguez: Why is precomputing the hash important?

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
```diff
@@ -71,19 +81,24 @@ static long arrayBaseOffset() {
...
+        return value.hashCode();
+    }
+
     @Snippet
-    public static String create(@ConstantParameter String compilationTimeString) {
-        int i = compilationTimeString.length();
+    public static String create(@ConstantParameter String value) {
+        int i = value.length();
         char[] array = (char[]) NewArrayNode.newUninitializedArray(char.class, i);
-        Word cArray = CStringConstant.cstring(compilationTimeString);
```
tkrodriguez: This is really constructing a modified UTF8 byte[] which you're just treating as ASCII.  Can't this be done in a cleaner way?  Using UNSAFE to build a String instance creates some problems with 9 since strings are more complicated there.  Couldn't we do something more layered where the snippet is only responsible for producing a UTF8 byte[] that is then consumed by a new String call in the original Java rootName method which is inlined normally?  It's extra allocation but it would all be much simpler.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/ControlFlowPathNode.java
```diff
@@ -61,11 +71,16 @@ private static boolean isCFGAccessible(FixedNode from, FixedNode to) {
...
+    public void preInlineInstrumentation(InstrumentationNode instrumentation) {
+    }
+
+    @Override
+    public void postInlineInstrumentation(InstrumentationNode instrumentation) {
         if (instrumentation.target() instanceof AbstractMergeNode) {
             AbstractMergeNode merge = (AbstractMergeNode) instrumentation.target();
-
-            if (isCFGAccessible(merge, position)) {
+            if (isCFGAccessible(merge, instrumentation)) { // ensure the scheduling
```
tkrodriguez: I can't quite parse this comment.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/InstrumentationBeginNode.java
```diff
@@ -20,43 +20,56 @@
...
     }
 
-    public int getOffset() {
+    public int offset() {
         return offset;
     }
 
-    public boolean inspectInvocation() {
-        return inspectInvocation;
+    public ValueNode target() {
```
tkrodriguez: Try to use get/set names, particularly when you have a set already.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/IsMethodInlinedNode.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public IsMethodInlinedNode(int depth) {
+        super(TYPE, StampFactory.forKind(JavaKind.Boolean));
+        this.parsingDepth = depth;
+    }
+
+    @Override
+    protected void afterClone(Node other) {
+        if (other instanceof IsMethodInlinedNode) {
+            // keep trace of the original graph
```
tkrodriguez: trace -> track

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/IsMethodInlinedNode.java
```diff
@@ -0,0 +1,89 @@
...
+            IsMethodInlinedNode that = (IsMethodInlinedNode) other;
+            this.originalGraph = that.originalGraph == null ? that.graph() : that.originalGraph;
+        }
+    }
+
+    private void resolve(StructuredGraph target) {
+        // if parsingDepth is greater than 0, then inlined
+        // if the origianl graph is null, which means this node is never cloned, then not inlined
```
tkrodriguez: original

graal/com.oracle.graal.virtual/src/com/oracle/graal/virtual/phases/ea/PartialEscapeClosure.java
```diff
@@ -181,6 +181,11 @@ private boolean processNodeInternal(Node node, BlockT state, GraphEffectList eff
                 return true;
             }
         }
+        if (UseGraalInstrumentation.getValue() && (node instanceof InstrumentationNode)) {
```
tkrodriguez: Why does InstrumentationNode need its own special case here?  Wouldn't it be properly handled by the following logic?  It's a ValueNode and it's Virtualizable.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
dougxc: This just made me realise that maybe this whole API is *only* intended for use in tools that do bytecode instrumentation. In this case, bytecode offsets make sense but then I'm wondering if you need offsets at all since you have precise control of where the calls to the API methods are placed. For example, couldn't you have something like:
```
invokestatic GraalDirectives.instrumentationBeginForSuccessor()
new java.lang.Object
invokestatic AllocationProfiler.countActualAllocation()
invokestatic GraalDirectives.instrumentationEnd()
invokespecial java.lang.Object()
```
You'd also have `GraalDirectives.instrumentationBegin()` for the case where instrumentation has no target instruction.

In any case, the javadoc should clearly state that these methods should not be used directly in Java source code. Also, the tests/examples should use bytecode assembly to align with and demonstrate the intended usage.

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -730,6 +811,48 @@ protected void build(FixedWithNextNode startInstruction, FrameStateBuilder start
...
+            if (node instanceof InstrumentationBeginNode) {
+                InstrumentationBeginNode begin = (InstrumentationBeginNode) node;
+                if (begin.offset() == 0) {
+                    // no further action for 0 offset
+                    return;
+                }
+                int targetIdx = -1;
+                int offset = begin.offset();
+                if (offset < 0) {
+                    // for a negative offset, we substrate the target index by 1 due to the input to
```
tkrodriguez: subtract

graal/com.oracle.graal.virtual/src/com/oracle/graal/virtual/phases/ea/PartialEscapeClosure.java
mur47x111: The following logic triggers only if there is any virtual input. In the case of InstrumentationNode, it needs a callback if its target node is monitorenter and the monitorenter is virtualized (when the monitorenter's object is virtualized).

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/ControlFlowPathNode.java
mur47x111: The ControlFlowPathNode indicates a dependency edge from a ValuePhiNode of this AbstractMergeNode to the instrumentation. The then-branch may trigger a schedule error if we don't put this guard.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
mur47x111: I agree that it is not necessary to precompute the hashcode as it will anyway be computed once.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
mur47x111: The previous discussion was how to pass a compilation-time String instance to the compiled code. The solution that we came out was to downgrade the String into a cstring using CStringConstant.

As long as UNSAFE.objectFieldOffset returns the correct offset, the UNSAFE writes to the newly created String will be "canonical" into StoreFieldNode (see UnsafeAccessNode.canonical).

I agree producing only UTF8 byte[] makes the code simpler. One issue is that we replace (instead of inline) the call to the original rootName method, which means it may not make a difference if we put it outside or in the snippet. I tried to directly put a new statement in the snippet (don't know if it is legal..) and it works. In this case, the snippet generates redundant array allocation -- it makes a copy of the char array in the String constructor. And that is the reason I discard it.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
mur47x111: Will update the javadoc and add more tests.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
mur47x111: At first, the whole API is intended for bytecode instrumentation. The offset was added for supporting source code where the target is hard to define. In this case, the user has to take care of counting the bytecode offset. If we agree on the support of source code is problematic, I will remove the offset mechanism.

Following your suggestion, I would prefer the target being excluded from the instrumentation, which is:
```
new java.lang.Object
invokestatic GraalDirectives.instrumentationBeginForPredecessor()
invokestatic AllocationProfiler.countActualAllocation()
invokestatic GraalDirectives.instrumentationEnd()
invokespecial java.lang.Object()
```
This makes it a bit easier for detaching the instrumentation from the original graph.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/RootNameNode.java
```diff
@@ -20,44 +20,56 @@
...
         super(TYPE, stamp);
     }
 
-    public static String genRootName(ResolvedJavaMethod method) {
-        if (method == null) {
-            return "<unresolved method>";
-        }
-        return method.getDeclaringClass().toJavaName() + "." + method.getName() + method.getSignature().toMethodDescriptor();
+    /**
+     * resolve this node and replace with a NewStringNode that constructs the root method name in
```
dougxc: `Resolves this node and replaces it with a {@link NewStringNode} that ...`

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/debug/instrumentation/RootNameNode.java
```diff
@@ -20,44 +20,56 @@
...
+     * the compiled code. To ensure the correct result, this method should be invoked only after the
+     * InliningPhase.
```
dougxc: `{@link InliningPhase}`

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
dougxc: The jdk9 issue Tom alludes to is [Compact Strings](http://openjdk.java.net/jeps/254) which means `String.value` is now a `byte[]`. As such, I second Tom's suggestion to generate a call to a `String` constructor.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
dougxc: Even better if we can remove the notion of a target from the instrumentation altogether - I didn't think that was possible.

Yes, we need to make it clear this mechanism is for bytecode instrumentation only. As such, can you please rework all the tests to use ASM (like [InterfaceMethodHandleTest](https://github.com/graalvm/graal-core/blob/master/graal/com.oracle.graal.compiler.test/src/com/oracle/graal/compiler/test/InterfaceMethodHandleTest.java#L147)) for example. It would also help in terms of demonstrating the purpose of each API method if the tests represent some kind of interesting application of the API.

mur47x111: I don't mean to remove the notion of the target, since it is still needed for relocating instrumentation in cases of optimizations like PEA. What I said is to put the target bytecode (new bytecode in the example) outside of the instrumentation scope marked by instrumentationBegin() and instrumentationEnd().

dougxc: Ok, that makes more sense. So, just to be clear, instrumentation for a target always uses `instrumentationBeginFor Predecessor` and non-targeted instrumentation uses `instrumentationBegin`.

BTW, what's an example of non-targeted instrumentation?

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
mur47x111: OK, then it is essential to update the snippet because my assumption was the fields did not change. Shall I use the String(byte[]) constructor? Also is it legal to directly put a new statement in the snippet?

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
mur47x111: For instance, if we still want to profile the conceptual allocation count, and don't want the instrumentation to perturb the inlining, we could mark the instrumentation non-targeted. By combining the aforementioned conceptual allocation count and the exact allocation count (via a targeted instrumentation), we get the stack allocation number.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
mur47x111: Or shall I make the directive API return a byte[]? The construction of the String is then in the custom code.

dougxc: You mean make `GraalDirective.rootName` return `byte[]`? I think the caller will always immediately convert this to a `String` so I don't think that's the right way to go. Calling `new String(byte[])` in the snippet should work.

graal/com.oracle.graal.virtual/src/com/oracle/graal/virtual/phases/ea/PartialEscapeClosure.java
tkrodriguez: Actually I think you want InstrumentationNode to implement VirtualizableAllocation.  The name is odd but it already has the semantics you want, so no changes would be required here.

mur47x111: Yes, you are right.. I will use this interface and remove the changes in PartialEscapeClosure.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewStringSnippets.java
tkrodriguez: Yeah I guess we'd like this to float and common since it's really a compile time constant, so if we can do it in the snippet that would be better.  new String(byte[]) uses the default platform charset not UTF8.  We'd really need to use new String(byte[], Charset) or new String(byte[], "UTF-8") but then we're back to having a dangling reference to a live Java object, which might cause the reference to get dropped.  Why not just store it as a 16 bit chars instead of as a cstring so we can directly construct it using new String(char[])?  If we had some support for strong references to objects from nmethods then we could just avoid this whole thing.

mur47x111: I have tried new String(byte[]) in the snippet, and it fails at the parsing phase because "All non-recursive calls in the intrinsic must be inlined or intrinsified". The new String(char[]) works fine. Have to check if it still works in JDK9 as the implementation is changed to compress it.

Isn't the compilation-time char array having the same issue of the dangling reference?

tkrodriguez: It wouldn't be a char[].  it would be just like your existing code to create a byte[] but the raw data we store would be a char instead of a byte.

I don't think we can safely call arbitrary String constructors inside a snippet.  It's just asking for trouble.  Maybe constructing a string for every call is the best way to go.  Actually RuntimeStringNode is a fixed node so it's not like it commons anyway.  It probably needs to be something like this where rawRootName is the piece that's intrinsified.

<pre>
    private static final Charset utf8 = Charset.forName("UTF-8");

    /**
     * @return the name of the root method for the current compilation task. If the enclosing method
     *         is inlined, it returns the name of the method into which it is inlined.
     */
    public static String rootName() {
        return new String(rawRootName(), utf8);
    }

    public static byte[] rawRootName() {
        return new byte[0];
    }
</pre>

mur47x111: I was also testing this. The issue is we have to force inlining rootName(), then we could substitute rawRootName().

tkrodriguez: Presumably an InlineInvokePlugin of some sort could do this.

mur47x111: Thanks. Will add one.

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -543,6 +549,71 @@ private void processPlaceholderFrameStates(IntrinsicContext intrinsic) {
         }
     }
 
+    /**
+     * This class maintains a mapping from BCI to a natural number sequence. This sequence is for
+     * indexing generated value nodes corresponding to the bytecodes indicated by the BCI.
+     */
+    private static class BCINodeMap {
+        int[] bci2Idx;
```
dougxc: Please use `index` instead of `idx`.

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -543,6 +549,71 @@ private void processPlaceholderFrameStates(IntrinsicContext intrinsic) {
...
+        int indexOf(ValueNode node) {
+            for (int i = 0; i < idxToNode.length; i++) {
+                if (idxToNode[i] == node) {
+                    return i;
+                }
+            }
+            return -1;
+        }
+
+        ValueNode get(int idx) {
```
dougxc: This method should be named `getNodeAt` or just `getNode`.

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -730,6 +811,48 @@ protected void build(FixedWithNextNode startInstruction, FrameStateBuilder start
         }
     }
 
+    private void resolveInstrumentationTarget() {
+        bciNodeMap.forEach(node -> {
```
dougxc: We discourage use of lambda in Graal given their startup cost (initializing the lambda system does a bunch of bytecode generation). At some point we will remove the existing use of lambdas in other parts of Graal.

In this case since `BCINodeMap` is a _friend_ inner class, you should just access `BCINodeMap.index2Node` directly and delete `BCINodeMap.forEach` (I didn't see any other uses).

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -543,6 +549,71 @@ private void processPlaceholderFrameStates(IntrinsicContext intrinsic) {
...
+            BytecodeStream stream = new BytecodeStream(method.getCode());
+            int idx = 0;
+            for (; stream.currentBC() != Bytecodes.END; stream.next()) {
+                bci2Idx[stream.currentBCI()] = idx++;
+            }
+            idxToNode = new ValueNode[idx];
+        }
+
+        /**
+         * Update generated value {@code node} corresponding to the bytecode indicated by
```
dougxc: `Sets the generated...`

From the javadoc [style guide](http://www.oracle.com/technetwork/articles/java/index-137868.html): 
```
Use 3rd person (descriptive) not 2nd person (prescriptive). 
The description is in 3rd person declarative rather than 2nd person imperative.

Gets the label. (preferred)

Get the label. (avoid)
```

graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -730,6 +811,48 @@ protected void build(FixedWithNextNode startInstruction, FrameStateBuilder start
         }
     }
 
+    private void resolveInstrumentationTarget() {
+        bciNodeMap.forEach(node -> {
+            if (node instanceof InstrumentationBeginNode) {
+                InstrumentationBeginNode begin = (InstrumentationBeginNode) node;
+                if (begin.getOffset() == 0) {
```
dougxc: I assume this will all change anyway now that we've agreed to remove `offset` from `GraalDirectives.instrumentationBegin...`.

graal/com.oracle.graal.api.directives.test/src/com/oracle/graal/api/directives/test/AllocationInstrumentationTest.java
```diff
@@ -31,90 +33,90 @@
...
-    private static class ClassA {
+    public static class ClassA {
         // This method should be marked as not inlineable
-        void notInlinedMethod() {
+        public void notInlinedMethod() {
         }
     }
 
-    static boolean flag;
+    public static boolean flag;
```
dougxc: If you rename this to `allocationWasExecuted`, it's easier to see how the test works.

graal/com.oracle.graal.api.directives.test/src/com/oracle/graal/api/directives/test/AllocationInstrumentationTest.java
```diff
@@ -31,90 +33,90 @@
...
 
     @Test
     public void testMustEscape() {
         try (OverrideScope s = OptionValue.override(GraalOptions.UseGraalInstrumentation, true)) {
-            ResolvedJavaMethod method = getResolvedJavaMethod("mustEscapeSnippet");
+            Class<?> clazz = instrumentor.instrument(AllocationInstrumentationTest.class, "mustEscapeSnippet", Opcodes.NEW);
+            ResolvedJavaMethod method = getResolvedJavaMethod(clazz, "mustEscapeSnippet");
             executeExpected(method, null); // ensure the method is fully resolved
             resetFlag();
             // The allocation in the snippet escapes. We expect the instrumentation preserves.
```
dougxc: `We expect the instrumentation is preserved.`

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
```diff
@@ -410,13 +410,21 @@ public static void instrumentationEnd() {
...
 
     /**
-     * @return an integer representing a control flow path taken for a @Snippet. This method is
-     *         valid only if invoked within an instrumentation (see
-     *         {@link #instrumentationBeginForPredecessor()} and {@link #instrumentationEnd()}), and
-     *         the associated target node of the instrumentation is a preceding node that will be
-     *         substituted by a @Snippet with multiple control flow paths. It will be replaced with
-     *         a ValuePhiNode with constant integer inputs [0, N-1], where N denotes the number of
-     *         control flow paths of the snippet.
+     * @return an integer representing a control flow path taken for allocation or synchronization
```
dougxc: This is the wrong place for describing paths related to platform and VM/runtime specific implementations of bytecodes. In addition, the values mentioned here are not guaranteed to hold if the relevant snippets are modified or the way the paths i the snippets are laid are changed to graph scheduler changes.

dougxc:
Can you please merge in the latest changes from upstream before too long.

graal/com.oracle.graal.api.directives/src/com/oracle/graal/api/directives/GraalDirectives.java
mur47x111: You are right. The initial idea is to get notified which path is taken in a snippet. The current implementation assigns each path an integer, but does not know about what the value represents. Any suggestion how to describe this? Or shall I simply drop this API method from the pull request and leave it researchy?

dougxc: I think you will need extra methods in `GraalDirectives` that can be used in snippets to associate control paths with meaningful names. I suggest we remove the concept from this PR and think further how that might look.

mur47x111: Sounds good to me.

gilles-duboscq:
Yudi, can you use your usi.ch email address or is it not valid anymore?

mur47x111:
I amended the author information.

gilles-duboscq:
Thank you :)

tkrodriguez:
Can you update to latest graal-core?  That should allow the gate to run properly.  I think all review comments have been addressed right?

mur47x111:
I have added SuppressWarnings. The newest error "com.oracle.graal.compiler.target.Backend.ARITHMETIC_EXP can not be resolved" seems exist in the trunk

mur47x111:
The bootstrap gate encounters a timeout. Any idea how to solve this?

tkrodriguez:
I think that's my fault.  I was rearranging the gate tags to make our internal gate work better but that also changed the github gate.  I'll fix this.

tkrodriguez:
sync with master and try again.  I'll run it through our internal gate too in preparation for pushing it.

mur47x111:
Please let me know if I could help with solving the issue. After fixing this, I will once again merge the upstream.

tkrodriguez:
I think it's unhappy about the fact that RootNameNode modifies memory after being lowered but before it's lowered there's no indication that it does that.  So the memory graph has incorrect edges.  I'm not sure why it doesn't complain in JDK8.  I'm looking into a fix.

tkrodriguez:
The new nodes need to be memory checkpoints.
Apply this change [fix.txt](https://github.com/graalvm/graal-core/files/385098/fix.txt) and sync with latest graal and I should be able to commit it.  There are merge conflicts that will have to be resolved.

mur47x111:
Seems like the timeout issue occurs again.

tkrodriguez:
Yes I'm seeing the timeout in another push but everything seems ok in our internal gate.  These travis machines are somewhat underpowered so maybe that's the problem.  I'll check the run times for that gate task on other machines.  Maybe we have a new expensive assert.

tkrodriguez:
Actually it looks like the bootstrap isn't terminating for some reason.  I think it might be a deopt pathology triggered by the slow gate.

PR #190, opened 2016-06-24
Generate overview of branch statistics, quick hack

smarr:
This is a quick hack to accumulate the an overview of that data collected. From my understanding, these are the interesting bits, but I might be wrong.

More work needed. That's merely a proposal to perhaps make the data a little more digestible.

With this change, the output is accumulated to produce entries like this:
```
at som.interpreter.nodes.dispatch.CachedDispatchNode.executeDispatch(CachedDispatchNode.java:38) [bci: 10]
NONE: 29 IF: 3 ELSE: 5 BOTH: 33
```

### DIFF
--- a/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/phases/InstrumentBranchesPhase.java
+++ b/graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/phases/InstrumentBranchesPhase.java
@@ -28,8 +28,10 @@ import static com.oracle.graal.truffle.TruffleCompilerOptions.TruffleInstrumentB
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
 import com.oracle.graal.compiler.common.type.StampFactory;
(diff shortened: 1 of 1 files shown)


axel22:
Looks good (and useful) to me! I recently pushed this commit in which I use execution counts per each if-branch, instead of just boolean flags:

https://github.com/graalvm/graal-core/commit/c3f892aba6d24a9fe9f2f5b0469e1dcc96ddad03

graal/com.oracle.graal.truffle/src/com/oracle/graal/truffle/phases/InstrumentBranchesPhase.java
```diff
@@ -144,8 +146,43 @@ private static String filterAndEncode(Node ifNode) {
             }
         }
 
+        private static class PointStats {
+            long NONE;
+            long IF;
+            long ELSE;
+            long BOTH;
+
+            public String toString() {
```
tkrodriguez: You need @Override here.

smarr: Yeah, the code also uses lambdas, so not sure this is really desirable in Graal.

tkrodriguez: The lambda's were already there so I wouldn't worry about that.  But it's definitely a case where the number of characters would be the same without lambdas, but it wouldn't be formatted as one line.  Lambdas seem to really lose on the conciseness front.

PR #192, opened 2016-07-01
Addition of libm sin, cos and tan plus assembler support. This code i…

mcberg2016:
…s OCA signed..


rschatz:
In `AMD64MathIntrinsicOp` you're adding new `temp` registers unconditionally. They are only used by the `sin`, `cos` and `tan` stubs, not by the `log` and `log10` stubs. I think they should only be allocated when they are really needed.

Other than that, looks good.

tkrodriguez:
There seem to be some unused declarations in AMD64MathIntrinsicOp.java.  I can't directly note them since github thinks the diff is too big.  mx gate -t BuildJavaWithEcj will show them to you.  You'll need to have ecj.jar available for that command to work.  You can download a copy from http://archive.eclipse.org/eclipse/downloads/drops4/R-4.5.2-201602121500/ecj-4.5.2.jar.  See the environment variable JDT to point at the downloaded file.

PR #195, opened 2016-07-09
use jvmci enabled jdk8 builds

dougxc:
Revives the Travis gate using the openjdk based jvmci binaries at https://github.com/dougxc/openjdk8-jvmci-builder/releases.


rschatz:
lgtm

Any idea why the bootstrap takes that much longer than the unit tests?

dougxc:
Bootstrap with system assertions enabled always takes longer and is probably amplified by a machine with less threads (the Travis trusty environment only provides 2 cores).

dougxc:
This change was made in another commit.

PR #199, opened 2016-07-28
addition of AVXop for binary operations to handle 3 address register …

mcberg2016:
…allocation

### DIFF
--- a/graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
+++ b/graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
@@ -719,7 +719,7 @@ public class AMD64Assembler extends Assembler {
     }
 
     /**
-     * Opcode with operand order of either RM or MR.
+     * Opcode with operand order of either RM or MR for 2 address forms.
      */
     public abstract static class AMD64RROp extends AMD64Op {
 
@@ -732,6 +732,22 @@ public class AMD64Assembler extends Assembler {
         }
 
         public abstract void emit(AMD64Assembler asm, OperandSize size, Register dst, Register src);
+    }
+
+    /**
+     * Opcode with operand order of either RM or MR for 3 address forms.
+     */
+    public abstract static class AMD64RRROp extends AMD64Op {
+
+        protected AMD64RRROp(String opcode, int prefix1, int prefix2, int op, OpAssertion assertion, CPUFeature feature) {
+            super(opcode, prefix1, prefix2, op, assertion, feature);
+        }
+
+        protected AMD64RRROp(String opcode, int prefix1, int prefix2, int op, boolean dstIsByte, boolean srcIsByte, OpAssertion assertion, CPUFeature feature) {
+            super(opcode, prefix1, prefix2, op, dstIsByte, srcIsByte, assertion, feature);
+        }
+
+        public abstract void emit(AMD64Assembler asm, OperandSize size, Register dst, Register nds, Register src);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


graal/com.oracle.graal.asm.amd64/src/com/oracle/graal/asm/amd64/AMD64Assembler.java
```diff
@@ -1227,16 +1365,129 @@ protected AMD64RMIOp(String opcode, boolean immIsByte, int prefix, int op, OpAss
 
         public final void emit(AMD64Assembler asm, OperandSize size, Register dst, Register src, int imm) {
             assert verify(asm, size, dst, src);
-            emitOpcode(asm, size, getRXB(dst, src), dst.encoding, src.encoding);
-            asm.emitModRM(dst, src);
-            emitImmediate(asm, size, imm);
+            boolean isSimd = false;
+            boolean noNds = false;
+
+            // TODO: add opcodes from the initializer here
```
tkrodriguez: Is there something to be done here?  Same with the one further down.

graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64Binary.java
```diff
@@ -92,10 +93,54 @@ public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
...
     /**
+     * Instruction that has three {@link AllocatableValue} operands.
+     */
+    public static class ThreeOp extends AMD64LIRInstruction {
+        public static final LIRInstructionClass<ThreeOp> TYPE = LIRInstructionClass.create(ThreeOp.class);
+
+        @Opcode private final AMD64RRMOp opcode;
+        private final OperandSize size;
+
+        @Def({REG, HINT}) protected AllocatableValue result;
```
tkrodriguez: You can drop the hint here I think though it shouldn't matter.

graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64Binary.java
```diff
@@ -92,10 +93,54 @@ public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
...
+        @Use({REG, STACK}) protected AllocatableValue x;
+        @Alive({REG, STACK}) protected AllocatableValue y;
```
tkrodriguez: Since this isn't commutative only y should allow stack I think.  Same for TwoOp.  We should let the allocator put the value into a register if it needs to be there.

tkrodriguez:
This is somewhat unrelated but having x and y both claim that they can be stack leads to some ugly looking generated code that probably doesn't perform well.  We can end up having each instruction have to load a stack slot into a register without being able to reuse it for later operations.

    0x00000001085e9aeb: vmovsd xmm1,QWORD PTR [rsp+0x28]
    0x00000001085e9af1: vmulsd xmm1,xmm1,QWORD PTR [rsp+0x20]
    0x00000001085e9af7: vmovsd xmm0,QWORD PTR [rsp+0x30]
    0x00000001085e9afd: vsubsd xmm0,xmm0,xmm1
    0x00000001085e9b01: vmovsd xmm3,QWORD PTR [rsp+0x38]
    0x00000001085e9b07: vmulsd xmm3,xmm3,QWORD PTR [rsp+0x30]
    0x00000001085e9b0d: vsubsd xmm0,xmm0,xmm3
    0x00000001085e9b11: vmovsd xmm1,QWORD PTR [rsp+0x30]
    0x00000001085e9b17: vmulsd xmm1,xmm1,QWORD PTR [rsp+0x20]
    0x00000001085e9b1d: vaddsd xmm1,xmm1,QWORD PTR [rsp+0x28]
    0x00000001085e9b23: vmovsd xmm2,QWORD PTR [rsp+0x28]
    0x00000001085e9b29: vmulsd xmm2,xmm2,QWORD PTR [rsp+0x38]
    0x00000001085e9b2f: vsubsd xmm1,xmm1,xmm2


rsp+0x20 and rsp+0x30 are both used multiple times with explicit reloads from the stack.  Some of this would be improved by removing STACK from the x argument of non-commutative arguments but commutative ones are still a problem.  We'd need something smarter in linear scan where it could ask an op if it can support a stack slot for an argument and convert to a different instruction in that case.  C2 does something like this is calls cisc spilling.

graal/com.oracle.graal.lir.amd64/src/com/oracle/graal/lir/amd64/AMD64Binary.java
```diff
@@ -118,20 +118,15 @@ public ThreeOp(AMD64RRMOp opcode, OperandSize size, AllocatableValue result, All
...
             } else {
-                AllocatableValue input = x;
+                //AllocatableValue input = x;
                 assert isStackSlot(y);
-                if (isStackSlot(x)) {
-                    input = result;
-                    AMD64Move.move(crb, masm, input, x);
-                }
-                opcode.emit(masm, size, asRegister(result), asRegister(input), (AMD64Address) crb.asAddress(y));
+                //if (isStackSlot(x)) {
```
tkrodriguez: I assume you meant to delete this completely?

mcberg2016: Yeah, its time is removed the "remind me to update this comment"

mcberg2016: will clean up

PR #211, opened 2016-10-07
minor code change to the pow intrinsic to correct a corner case

mcberg2016:
This corrects a corner case posed in the Math.pow unittest.


dougxc:
Is it possible to write a unit test for this (i.e., something that would fail without the change)?

tkrodriguez:
I added a unit test that shows this problem when I disabled the pow intrinsic.  So it's already being tested.

PR #214, opened 2016-10-15
source position in inlined nodes is now set correctly

mur47x111:
putIfAbsent returns null for new entry and thus results in empty source position


tkrodriguez:
Thanks for the fix!

PR #215, opened 2016-11-02
Prepare for lowering to expose some constants early

veresov:
This is a small part of the AOT changes just to test the process with GitHub out. The proposed change introduces a new API point `prepareForLowering()` in `Lowerable` in order to allow nodes to prepare to the lowering process. In AOT this is used to expose some of the constants for nodes that would otherwise be lowered after the frame states are fixed, which is too late for the constant replacement phase (will be introduced later) to work with.

### DIFF
--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/arraycopy/ArrayCopySlowPathNode.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/arraycopy/ArrayCopySlowPathNode.java
@@ -29,6 +29,7 @@ import static com.oracle.graal.compiler.common.LocationIdentity.any;
 
 import com.oracle.graal.compiler.common.LocationIdentity;
 import com.oracle.graal.graph.NodeClass;
+import com.oracle.graal.hotspot.word.KlassPointer;
 import com.oracle.graal.nodeinfo.InputType;
 import com.oracle.graal.nodeinfo.NodeInfo;
 import com.oracle.graal.nodes.NamedLocationIdentity;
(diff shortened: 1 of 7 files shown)


dougxc:
The timeout in the Travis gate is due to an infinite loop bug in `DefaultLoopPolicies` that I'm pushing a fix for now. Once it lands on github you'll need to update/rebase this PR.

dougxc:
The timeout fix is in https://github.com/graalvm/graal-core/commit/efb8df47ec55ad5a9c67734ac0b8162e7101cb19

tkrodriguez:
I'm not sure I get this change.  prepareForLowering it called right before lower so it seems like it doesn't gain you anything unless they are done in different lowering phases.  What you're really trying to accomplish is to capture some information in an earlier lowering phase for use in a later lowering right?  Why can't you simply do that when lower is called in DefaultHotSpotLoweringProvider?  There's already some funny setMirror logic for InstanceOfDynamicNode for instance and that seems like where the prepareForLowering logic should live.

veresov:
I guess it felt more structured at the time. But I can do it similarly to setMirror, sure.

veresov:
Alright, I've updated the change accordingly.

graal/com.oracle.graal.nodes/src/com/oracle/graal/nodes/java/DynamicNewArrayNode.java
```diff
@@ -48,6 +48,11 @@
 
     @Input ValueNode elementType;
 
+    // Class pointer to void.class needs to be exposed earlier than this node is lowered so that it
+    // can be replaced by the AOT machinery. If it's not needed for lowering this input can be
+    // ignored.
+    @OptionalInput ValueNode voidClass;
```
dougxc: This should be a javadoc style comment. Same applies in `DynamicNewInstanceNode.java`.

graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/arraycopy/ArrayCopySlowPathNode.java
```diff
@@ -49,16 +50,26 @@
      */
     private final Object argument;
 
-    public ArrayCopySlowPathNode(ValueNode src, ValueNode srcPos, ValueNode dest, ValueNode destPos, ValueNode length, JavaKind elementKind, SnippetTemplate.SnippetInfo snippet, Object argument) {
+    /**
+     * AOT compilation requires klass constants to be exposed after the first lowering to be handled
+     * automatically. Lowering for ArrayCopySlowPathNode, with snippet ==
```
dougxc: `{@link ArrayCopySlowPathNode}`
`{@link ArrayCopySnippets#arraycopyPredictedObjectWork}`

PR #218, opened 2016-11-12
Expand Instanceof at high tier

mur47x111:
InstanceOfNode are expanded to a potential null check and a LoweredInstanceOfNode that test on a non-null object and without profile. The ultimate goal is to get rid of ShortCircuitOrNode.


graal/com.oracle.graal.phases.common/src/com/oracle/graal/phases/common/ExpandInstanceOfPhase.java
```diff
@@ -0,0 +1,473 @@
...
+ * subgraph will be inserted before the {@code FixedNode} or the anchor. Otherwise, the subgrpah
+ * will be inserted at the latest {@code FixedNode} in the current schedule.
+ */
+public class ExpandInstanceOfPhase extends BasePhase<PhaseContext> {
+
+    @Override
+    protected void run(StructuredGraph graph, PhaseContext context) {
+        // The first SchedulePhase is for generating loop exit nodes for transforming fixed guard
+        // within loops.
+        new SchedulePhase(SchedulingStrategy.LATEST).apply(graph, false);
```
tkrodriguez: Two extra SchedulePhases seems incredibly expensive.  Does it have to be done this?

mur47x111:
Close until we have strong evidence of benefit.

PR #219, opened 2016-11-12
ControlFlowAnchorNode should have prevented LoopUnswitchingPhase

mur47x111:
Besides, `GraalCompilerTest` inserts a `ComputeLoopFrequenciesPhase` after loop optimizations, making `GraalDirectives.injectIterationCount` useless.


graal/com.oracle.graal.loop/src/com/oracle/graal/loop/DefaultLoopPolicies.java
```diff
@@ -183,7 +183,23 @@ public boolean shouldUnswitch(LoopEx loop, List<ControlSplitNode> controlSplits)
 
         Debug.log("shouldUnswitch(%s, %s) : delta=%d (%.2f%% inside of branches), max=%d, f=%.2f, phis=%d -> %b", loop, controlSplits, actualDiff, (double) (inBranchTotal) / loopTotal * 100, maxDiff,
                         loopFrequency, phis, actualDiff <= maxDiff);
-        return actualDiff <= maxDiff;
```
mur47x111: ok, does this also apply to `DefaultLoopPolicies.shouldPeel` and `DefaultLoopPolicies.shouldFullUnroll`?

graal/com.oracle.graal.compiler.test/src/com/oracle/graal/compiler/test/GraalCompilerTest.java
```diff
@@ -185,7 +185,7 @@ protected Suites createSuites() {
         Suites ret = backend.getSuites().getDefaultSuites().copy();
         ListIterator<BasePhase<? super HighTierContext>> iter = ret.getHighTier().findPhase(ConvertDeoptimizeToGuardPhase.class, true);
         if (iter != null) {
-            PhaseSuite.findNextPhase(iter, CanonicalizerPhase.class);
+            // the ComputeLoopFrequenciesPhase will be inserted before LoopFullUnrollPhase
```
mur47x111: In the economy configuration we don't have the ConvertDeoptimizeToGuard phase so the fix won't affect the economy configuration. We will have
```
        if (iter == null) {
            iter = ret.getHighTier().findPhase(CanonicalizerPhase.class);
        }
```
for economy configuration case.

graal/com.oracle.graal.loop/src/com/oracle/graal/loop/DefaultLoopPolicies.java
tkrodriguez: Oh I see, you're replicating logic that already has 2 copies lying around.  Maybe factor out that logic into LoopEx.canDuplicateLoop and call that instead.  @gilles-duboscq  and @rschatz does that sound right?  And should we have asserts in the loop fragments that we aren't duplicating loops with these constraints?

What's the reason for the FrameState restrictions?

tkrodriguez:
I had to make some minor changes to the unswitch test to make it pass in all our repos.

PR #221, opened 2016-11-14
Constructors with final fields need a release memory barrier.

theRealAph:
I know this is incomplete: there's nothing to make sure that the barrier works when the return expands inline to a GOTO, for example. And I don't think that there's anything to stop the Membar from being moved. I have spent a little time trying to figure out a really elegant way to solve the problem, but I can't see one.

### DIFF
--- a/graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
+++ b/graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
@@ -240,6 +240,8 @@ import static com.oracle.graal.java.BytecodeParserOptions.UseGuardedIntrinsics;
 import static com.oracle.graal.nodes.graphbuilderconf.IntrinsicContext.CompilationContext.INLINE_DURING_PARSING;
 import static com.oracle.graal.nodes.type.StampTool.isPointerNonNull;
 import static java.lang.String.format;
+import static jdk.vm.ci.code.MemoryBarriers.LOAD_STORE;
+import static jdk.vm.ci.code.MemoryBarriers.STORE_STORE;
 import static jdk.vm.ci.meta.DeoptimizationAction.InvalidateRecompile;
 import static jdk.vm.ci.meta.DeoptimizationAction.InvalidateReprofile;
 import static jdk.vm.ci.meta.DeoptimizationReason.JavaSubroutineMismatch;
@@ -366,6 +368,7 @@ import com.oracle.graal.nodes.extended.GuardingNode;
 import com.oracle.graal.nodes.extended.IntegerSwitchNode;
 import com.oracle.graal.nodes.extended.LoadHubNode;
 import com.oracle.graal.nodes.extended.LoadMethodNode;
+import com.oracle.graal.nodes.extended.MembarNode;
 import com.oracle.graal.nodes.extended.ValueAnchorNode;
 import com.oracle.graal.nodes.graphbuilderconf.GraphBuilderConfiguration;
 import com.oracle.graal.nodes.graphbuilderconf.GraphBuilderConfiguration.BytecodeExceptionMode;
@@ -594,6 +597,8 @@ public class BytecodeParser implements GraphBuilderContext {
 
     private int lastBCI; // BCI of lastInstr. This field is for resolving instrumentation target.
 
+    private boolean needsReturnMembar = false; // Constructor has write to final field.
+
     protected BytecodeParser(GraphBuilderPhase.Instance graphBuilderInstance, StructuredGraph graph, BytecodeParser parent, ResolvedJavaMethod method,
                     int entryBCI, IntrinsicContext intrinsicContext) {
         this.bytecodeProvider = intrinsicContext == null ? new ResolvedJavaMethodBytecodeProvider() : intrinsicContext.getBytecodeProvider();
@@ -614,6 +619,7 @@ public class BytecodeParser implements GraphBuilderContext {
         this.entryBCI = entryBCI;
         this.parent = parent;
         this.lastBCI = -1;
+        this.needsReturnMembar = false;
 
         assert code.getCode() != null : "method must contain bytecodes: " + method;
 
@@ -1229,6 +1235,9 @@ public class BytecodeParser implements GraphBuilderContext {
         StoreFieldNode storeFieldNode = new StoreFieldNode(receiver, field, value);
         append(storeFieldNode);
         storeFieldNode.setStateAfter(this.createFrameState(stream.nextBCI(), storeFieldNode));
+        if (field.isFinal() && method.isConstructor()) {
+            needsReturnMembar = true;
+        }
     }
 
     /**
@@ -1989,6 +1998,11 @@ public class BytecodeParser implements GraphBuilderContext {
     }
 
     private void beforeReturn(ValueNode x, JavaKind kind) {
+        if (needsReturnMembar) {
+            FixedNode membarInstr = new MembarNode(LOAD_STORE | STORE_STORE);
+            append(membarInstr);
+        }
+
         if (graph.method() != null && graph.method().isJavaLangObjectInit()) {
             /*
              * Get the receiver from the initial state since bytecode rewriting could do arbitrary


graal/com.oracle.graal.java/src/com/oracle/graal/java/BytecodeParser.java
```diff
@@ -1989,6 +1998,11 @@ protected void genReturn(ValueNode returnVal, JavaKind returnKind) {
     }
 
     private void beforeReturn(ValueNode x, JavaKind kind) {
+        if (needsReturnMembar) {
+            FixedNode membarInstr = new MembarNode(LOAD_STORE | STORE_STORE);
```
eregon: The barrier should also only be emitted if the CPU does not already guarantee `STORE_STORE` ordering (both Intel and SPARC do).

theRealAph:
On 14/11/16 22:21, Benoit Daloze wrote:
> eregon commented on this pull request.
> 
> 
> 
>> @@ -1989,6 +1998,11 @@ protected void genReturn(ValueNode returnVal, JavaKind returnKind) {
>      }
>  
>      private void beforeReturn(ValueNode x, JavaKind kind) {
> +        if (needsReturnMembar) {
> +            FixedNode membarInstr = new MembarNode(LOAD_STORE | STORE_STORE);
> 
> I think only the `STORE_STORE` barrier is needed, because on the
> Load side, most CPUs consider data dependencies with indirection
> ordering (and I believe all recent CPUs do).  See "Data Dependency
> and Barriers" in http://g.oswego.edu/dl/jmm/cookbook.html.

We've been over this (at very great length) in the HotSpot list, and
come to the opposite conclusion.  The explanation is here:

http://www.hboehm.info/c++mm/no_write_fences.html

In summary: it is probably safe to use only a StoreStore fence if the
object is immutable, but even then it's not clear.  It's certainly not
worth the risk IMO.

Andrew.

theRealAph:
On 14/11/16 22:26, Benoit Daloze wrote:
> eregon commented on this pull request.
> 
>> @@ -1989,6 +1998,11 @@ protected void genReturn(ValueNode returnVal, JavaKind returnKind) {
>      }
>  
>      private void beforeReturn(ValueNode x, JavaKind kind) {
> +        if (needsReturnMembar) {
> +            FixedNode membarInstr = new MembarNode(LOAD_STORE | STORE_STORE);
> 
> The barrier should also only be emitted if the CPU does not already guarantee `STORE_STORE` ordering (both Intel and SPARC do).

That's already handled by the code which lowers the MembarNode.  However,
some sort of CPU barrier is still required to ensure that the compiler
doesn't reorder prior loads and stores.

Andrew.

dougxc:
@lukasstadler , the unit test [failure](https://travis-ci.org/graalvm/graal-core/jobs/175799456#L764) caused by these changes raises the question of how escape analysis needs to handle the introduction of membars in the high level graph. @gilles-duboscq was suggesting that we probably need to add something like a fixed `PublishNode` that is emitted just prior to a `ReturnNode` in a constructor that assigns final fields. This means EA can emit the right barrier when (if) it commits an allocation, possibly coalescing barriers.

On a related note, how does EA handle the publishing semantics for `volatile` fields assigned in the constructor of an EA'ed object?

tkrodriguez:
Regarding volatiles and EA, MemBarNode isn't virtualizable so we won't EA objects with volatile updates.  The initialization of volatile fields to 0 doesn't have any barrier in the normal case so as long as we don't see any other updates we will happily EA that and then lower it back down into a normal Java allocation.

tkrodriguez:
So in the case where STORE_STORE isn't required what kind of compiler reorderings are disallowed for final fields?  Graal already maintains all stores in the original program order so I'm not sure we need a membar to keep the compiler from reordering operations.

theRealAph:
On 15/11/16 16:39, Tom Rodriguez wrote:

> So in the case where STORE_STORE isn't required what kind of
> compiler reorderings are disallowed for final fields?  Graal already
> maintains all stores in the original program order so I'm not sure
> we need a membar to keep the compiler from reordering operations.

OK, I'm learning things all the time.  I had no idea that Graal didn't
reorder stores.

tkrodriguez:
I suspect we need MemBarNode to have more information about why it exists so it can play nice with EA.  It's my recollection that synchronized and barrier operations on objects which are known to be unreachable by other threads have no effect.  Can anyone confirm that?

So if MemBarNode tracked the node for which it was emitted then we could virtualize them and reemit them if we materialize through a CommitAllocation.  The only question I have there is whether we need multiple barriers and/or need to maintain the original store ordering.  It seems like we could simply perform all the writes for the group of objects we're allocating and then emit the required barrier instruction.

tkrodriguez:
Yes, graal uses a mix of fixed and floating nodes and currently stores are always fixed in the graph.  Reads start out fixed and can float like C2 does, but not all reads are floated and a floating read can turn back into a fixed read.

tkrodriguez:
Andrew, one thing implied in my earlier comments that's maybe not clear is that when Graal performs escape analysis it can pick up an object and all the writes to it and then later materialize all those operations.  It will even group multiple objects and do a single allocation and initialize them all at once.  That will also need the final barrier but since the existence of MemBarNode inhibits EA it's not actually a problem in your changes.  But we'll have to resolve the EA problem before we can accept the fix.

gilles-duboscq:
> Graal already maintains all stores in the original program order so I'm not sure we need a membar to keep the compiler from reordering operations.
That's not entirely true, after FrameState assignment, Graal can re-order stores.

theRealAph:
On 15/11/16 16:48, Tom Rodriguez wrote:
> It's my recollection that synchronized and barrier operations on objects which are known to be unreachable by other threads have no effect.  Can anyone confirm that?

That's right.  Explicit barriers must of course be retained but stores to volatiles that aren't loaded don't have any synchronization effect because there's nothing for them to synchronize with.

gilles-duboscq:
Also, the `MemBarNode` does not consume the receiver object so unless there would need to be a special effect because of the memory kill or special treatment of `MemBar` i don't think EA will care.
I couldn't find any such special effect or treatment.
I think EA didn't really need to care too much because because it runs before the lowering where the `MemBars` for volatile accesses are inserted and anyway, by its nature EA deals with objects that have not been published yet.
I think we should still insert some barriers when lowering the `CommitAllocationNode` in case some of the fields are final or volatile.

theRealAph:
On 15/11/16 09:27, Andrew Haley wrote:
> On 14/11/16 22:21, Benoit Daloze wrote:
>> eregon commented on this pull request.
>>
>> I think only the `STORE_STORE` barrier is needed, because on the
>> Load side, most CPUs consider data dependencies with indirection
>> ordering (and I believe all recent CPUs do).  See "Data Dependency
>> and Barriers" in http://g.oswego.edu/dl/jmm/cookbook.html.
> 
> We've been over this (at very great length) in the HotSpot list, and
> come to the opposite conclusion.

I've just discussed this (again :-) with Hans Boehm and Doug Lea.
Everybody agrees that STORE_STORE is adequate for this purpose: you're
right.

eregon:
> I've just discussed this (again :-) with Hans Boehm and Doug Lea.
Everybody agrees that STORE_STORE is adequate for this purpose: you're
right.

@theRealAph  Interesting, do you have any link to one of these discussions? Is it on concurrency-interest?

theRealAph:
On 16/11/16 10:28, Benoit Daloze wrote:

> @theRealAph  Interesting, do you have any link to one of these discussions? Is it on concurrency-interest?

Summary: StoreStore is adequate for the strict specification, but it might
surprise some programmers.

http://mail.openjdk.java.net/pipermail/jmm-dev/2016-November/000379.html

tkrodriguez:
We discussed how we think we should resolve the interactions with escape analysis and basically we need to introduce a new node to represent the end of the constructor and then this will later be lowered into the MemBarNode after EA, which will match the way that volatile field operations are done.  It will also add any needed barrier after the CommitAllocationNode.  I'll create a new github PR so you can confirm that it works for you.

By the way, are there any particular tests which were failing with the missing membar or was it just general instability?  I'd like to add a more directed test of this problem to make sure we also exercise the EA version of this problem.

theRealAph:
On 16/11/16 17:07, Tom Rodriguez wrote:

> By the way, are there any particular tests which were failing with
> the missing membar or was it just general instability?  I'd like to
> add a more directed test of this problem to make sure we also
> exercise the EA version of this problem.

It was general instability, although I'm sure that jcstress would reveal the
failures.

By the way, I've found more bugs.  I was surprised to discover that
there's no StoreStore barrier after object initialization either.
This means that other threads see stale data from previous
allocations.  (Argh!)  The fix for this one is pretty simple, as
appended.

I'm not sure what the protocol is here.  I can prepare a new patch
with a new branch, but these rather go together.

Andrew.


--- a/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewObjectSnippets.java
+++ b/graal/com.oracle.graal.hotspot/src/com/oracle/graal/hotspot/replacements/NewObjectSnippets.java
@@ -91,6 +91,7 @@ import com.oracle.graal.nodes.debug.DynamicCounterNode;
 import com.oracle.graal.nodes.debug.VerifyHeapNode;
 import com.oracle.graal.nodes.extended.BranchProbabilityNode;
 import com.oracle.graal.nodes.extended.ForeignCallNode;
+import com.oracle.graal.nodes.extended.MembarNode;
 import com.oracle.graal.nodes.java.DynamicNewArrayNode;
 import com.oracle.graal.nodes.java.DynamicNewInstanceNode;
 import com.oracle.graal.nodes.java.NewArrayNode;
@@ -115,6 +116,7 @@ import com.oracle.graal.word.Word;
 import jdk.vm.ci.code.CodeUtil;
 import jdk.vm.ci.code.Register;
 import jdk.vm.ci.code.TargetDescription;
+import static jdk.vm.ci.code.MemoryBarriers.STORE_STORE;
 import jdk.vm.ci.hotspot.HotSpotJVMCIRuntimeProvider;
 import jdk.vm.ci.hotspot.HotSpotResolvedObjectType;
 import jdk.vm.ci.meta.DeoptimizationAction;
@@ -447,6 +449,7 @@ public class NewObjectSnippets implements Snippets {
         } else if (REPLACEMENTS_ASSERTIONS_ENABLED) {
             fillWithGarbage(size, memory, constantSize, instanceHeaderSize(INJECTED_VMCONFIG), false, useSnippetCounters);
         }
+        MembarNode.memoryBarrier(STORE_STORE);
         return memory.toObject();
     }

@@ -478,6 +481,7 @@ public class NewObjectSnippets implements Snippets {
         } else if (REPLACEMENTS_ASSERTIONS_ENABLED) {
             fillWithGarbage(allocationSize, memory, false, headerSize, maybeUnroll, useSnippetCounters);
         }
+        MembarNode.memoryBarrier(STORE_STORE);
         return memory.toObject();
     }

theRealAph:
Can someone please comment on this? Aarch64 breaks without it and I think the fix is correct, but there is a weird test failure.

tkrodriguez:
Sorry I didn't update you.  I'm preparing a fix for this since the overall issue requires some larger changes related to escape analysis.  I'll include the barriers you've added as well.  I should have bits up for you to try in a day or so.

tkrodriguez:
for some reason I don't understand, github closed my PR with my version of this fix but I pushed my fix through our internal bitbucket as 70e29ac.  Can you try out the new bits and let me know if there are any issues with them?

smarr:
@tkrodriguez github saw that the commit hash matched. It then assumed that the PR is to be closed because the requested merge happened.

tkrodriguez:
@smarr No it was closed before I committed it to graal-core.  I created the PR and then a few hours later it was just closed.

PR #231, opened 2016-11-29
Add missing barriers for allocations and final fields

tkrodriguez:
This adds the required barriers for final fields and allocations.  The final field barriers can be EA'd and the lowered later into the proper barriers when the allocation is finally committed.


tkrodriguez:
@theRealAph So can you review this and confirm that it resolves the issues in #221?

PR #224, opened 2016-11-24
Many small AArch64 fixes

theRealAph:
These fix a bunch of minor problems in the AArch64 back end. They are impossible to test in isolation; they are the minimum needed to run a any tests.

They are:

Tableswitch completely rewritten. It generated incorrect code and did not properly construct a jump table.

Narrow OOPs were not supported at all. All done.

All memory loads sign extended. Rewritten to use either zero- or sign-extension depending on context.

Sign and zero extension rewritten in many places.

Bit scan operations returned an incorrect type, causing compiler errors.

CompareAndSwap returned a value instead of a boolean.

ByteSwap was unimplemented.

PatchReturnAddressOp did not work at all.

AArch64HotSpotSafepointOp did not use the correct relocation.

Temporary register usage in stack spills was incorrect. Stack-stack spills were not implemented at all.

Loading a stack address corrupted the stack pointer.

PR #240, opened 2016-12-09
The real aph aarch64 graal misc fixes

dougxc:


gilles-duboscq:
This is already merged in 45e1839 (see #224).

PR #241, opened 2016-12-17
Fix argument types of tiered callbacks

veresov:
A small change to straighten argument types in tiered callbacks.


veresov:
https://bugs.openjdk.java.net/browse/JDK-8171394

PR #243, opened 2017-01-05
Fix AOT unverified entry point

veresov:
Make references to oop base and klass base dynamic.


tkrodriguez:
I had to adjust the commit message so this won't close automatically.  This was pushed as https://github.com/graalvm/graal-core/commit/73d6f9ba1d2c5b5d1ee86776520935ec02955129.

PR #245, opened 2017-01-21
Avoid zero-shift with AOT

veresov:
AOT captures VM settings during compilation. For compressed oops it presents a problem for the case when VM selects a zero-shift mode (for heaps less than 4G). Compiling AOT binary with zero-shift limits it's usability. The AOT compiler should be able to avoid zero-shift.


dougxc:
Looks good to me. I'll integrate it.

veresov:
Thanks, @dougxc !

dougxc:
Fixed by https://github.com/graalvm/graal-core/commit/23acd8b195a3ccbe8ef080c4690567afc9b1fe8c

PR #248, opened 2017-01-30
Add support for Truffle on Aarch64.

theRealAph:
Truffle NFI support for AArch64.


dougxc:
@rschatz , I know you've said we should delete the NFI tests given the NFI rework you're currently doing. Maybe now is a good time to do it?

rschatz:
I'm currently working on a replacement for the old NFI, but this is mostly interface work on the Truffle side. There will be an intrinsification of this interface on the Graal side, but I haven't started with that yet.
The intrinsification of the new NFI will probably re-use most of the implementation code of the old NFI, just the public interface will be gone.

So I think it's still worthwile to accept this PR, for the backend code in there.

Regarding the tests: If they are not causing any major problems, I'd prefer them to stay at least until we have the intrinsification of the new NFI in place. The reason is that otherwise the implementation code (e.g. the `RawNativeCallNode`) will be completely untested...

dougxc:
Ok, sounds reasonable. Can you please then work with Andrew to merge this PR.

graal/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
```diff
@@ -486,7 +486,7 @@ public long tlabIntArrayMarkWord() {
     public final int methodCompiledEntryOffset = getFieldOffset("Method::_from_compiled_entry", Integer.class, "address");
     public final int methodCodeOffset = getFieldOffset("Method::_code", Integer.class, isJDK8 ? "nmethod*" : "CompiledMethod*");
 
-    public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
+    // public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
```
rschatz: Why did you comment this out?
In general we try not to have any commented code in the repo. That's what git history is for ;)

@dougxc Do you know why this flag is unused? Can/should we remove it?

log
```diff
@@ -0,0 +1,15 @@
+Compiling com.oracle.truffle.api.debug.test with javac-daemon... [/nfs/zebedee/home/graal/aph/truffle/mxbuild/truffle/com.oracle.truffle.api.debug.test/bin/com/oracle/truffle/api/debug/test/LegacyBreakpointTest.class[2017-01-25 17:18:29] is older than /nfs/zebedee/home/graal/aph/truffle/mxbuild/truffle/com.oracle.truffle.api.instrumentation.test/bin/META-INF/services/java.nio.file.spi.FileTypeDetector[2017-01-30 11:48:27]]
```
rschatz: Looks like accidentially added file.

graal/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
dougxc: http://mail.openjdk.java.net/pipermail/graal-dev/2017-January/004821.html

theRealAph:
On 30/01/17 15:53, Roland Schatz wrote:
>  
> -    public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
> +    // public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
> 
> Why did you comment this out?

Sorry, should not have been included but Graal doesn't build without that change.

Andrew.

theRealAph:
On 30/01/17 16:27, Andrew Haley wrote:
> On 30/01/17 15:53, Roland Schatz wrote:
>>  
>> -    public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
>> +    // public final int methodFlagsJfrTowrite = getConstant("Method::_jfr_towrite", Integer.class);
>>
>> Why did you comment this out?

Do you want me to back this change out?

Andrew.

rschatz:
Yes, please remove the change from this PR. We will probably remove this field soon anyway.

rschatz:
Update: The field was already removed by @dougxc , and now this PR conflicts because of this. If you just rebase, that issue should resolve itself...

log
theRealAph: What file is accidentally added? I have no idea.

theRealAph: I think this is now fixed.

rschatz:
I'm pretty sure the gate failure in travis is not your fault. Since it's not failing in our internal gates, it's very likely some other change is responsible for it and we just didn't notice. I consider this ready to merge.

PR #249, opened 2017-01-31
Fix types in GraalHotSpotVMConfigNode.inlineContiguousAllocationSupported

veresov:
In GraalHotSpotVMConfigNode.inlineContiguousAllocationSupported() we need a cast because loadConfigValue() returns a long, but the node it produces has an int type. It triggers an assert because of the types mismatch.


veresov:
Hm, it seems to be failing a truffle test with jdk9. Is it something that's normal or is my change?

dougxc:
You can ignore jdk9 gate failures for now as we have disabled internal gating on jdk9 until https://bugs.openjdk.java.net/browse/JDK-8172850 is in an Early Access build.

veresov:
Ok, I see, thanks! I've made a few more changes to this fix. Got rid of the fragile casts completely. And added some comments.

gilles-duboscq:
The backend uses the word size when generating LIR for this node.
If the backend is correct then `GraalHotSpotVMConfigNode` should be word-sized and the intrinsics should return `Word`. Otherwise we should fix the backend.

veresov:
It is a `Word` in the backend, yes. But the users of the` GraalHotSpotVMConfigNode` expect it returning longs, ints and booleans depending on the configuration parameter they need. So while the storage for any of those is 64bit on x64 these should be truncated to represent smaller values.  It is wrong how I do it now? What would be the right way?

tkrodriguez:
The backend and the frontend should agree.  So emitLoadConfigValue should take the expected kind and drive the code generation from that instead of always producing a Word for something that might expect an int or boolean.

veresov:
Tom, Gilles, is that closer to what you guys had in mind?

graal/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/GraalHotSpotVMConfigNode.java
```diff
@@ -55,62 +55,73 @@
...
     }
 
     public static int logOfHeapRegionGrainBytes() {
-        return (int) loadConfigValue(logOfHeapRegionGrainBytesMark(INJECTED_VMCONFIG), JavaKind.Byte);
+        return loadIntConfigValue(logOfHeapRegionGrainBytesMark(INJECTED_VMCONFIG), JavaKind.Int);
     }
 
     public static boolean inlineContiguousAllocationSupported() {
-        return loadConfigValue(inlineContiguousAllocationSupportedMark(INJECTED_VMCONFIG), JavaKind.Byte) > 0;
+        return loadIntConfigValue(inlineContiguousAllocationSupportedMark(INJECTED_VMCONFIG), JavaKind.Byte) != 0;
```
tkrodriguez: Why is this still JavaKind.Byte?  I would expect the JavaKind to be implied by the loadInt or loadLong part instead of separately specified.  Or is there something going on that I'm missing?

tkrodriguez:
That looks more consistent.

graal/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/GraalHotSpotVMConfigNode.java
veresov: Well, these are node intrinsics and they dispatch to the same constructor. Is there a way to the type in some less visible way?

veresov: I made loadLong and loadInt mostly to get the right return types.

tkrodriguez: I guess my point is that they should agree all the way through.  I think you want to use setStampFromReturnType = true in your NodeIntrinsic and then set the initial value as StampFactory.forNodeIntrinsic().  Construction of the node intrinsic will then override the stamp from the return type of the loadIntConfigValue and loadLoadConfigValue.

veresov: Hm, ok, let me try that..

veresov:
Alright, I've made it derive stamps from return types.

gilles-duboscq:
Looks good now. Thanks for the changes.

veresov:
Thanks Gilles! So, is it good to go?

tkrodriguez:
Yes, I'll create the bitbucket PR for this.

tkrodriguez:
Could you go ahead and rebase onto core?  There are some CI changes I think we need.

veresov:
Hm, ok, I refreshed it, but there didn't seem to be any additional changes required. What changes did you have in mind?

tkrodriguez:
No, I meant that we needed to sync with changes to the CI by rebasing to the current head.  I could do it myself but it's leaves cleaner github state if it's done here.

veresov:
I'm afraid I don't quite understand which repo should be the base.

tkrodriguez:
You've already done what I asked.  It just needed to be updated to latest graal-core.

veresov:
Ok, cool!

PR #250, opened 2017-02-02
Investigate travis failure

tkrodriguez:


dougxc:
@gilles-duboscq saw a Truffle failure today while making Graal work again with JDK 9. I think it may be the same one that @theRealAph reported and that I assume you're trying to reproduce here.

tkrodriguez:
Yes it seems pretty reliable in the gate https://travis-ci.org/graalvm/graal-core/jobs/197749872

tkrodriguez:
I was concerned that indexOf might be implicated but disabling it had no effect.  I'll try backing up a few commits instead.

tkrodriguez:
I'm investigating this over at https://github.com/tkrodriguez/graal-core/pull/1 since the gate automerges against master which defeats the ability to test individual changesets.

PR #266, opened 2017-04-03
Implement better address lowering for AArch64

adinn:
This change implements address lowering by use for AArch64 using a CPU-specific phase AddressLoweringByUsePhase which invokes a CPU-specific implementation AArch64AddressLoweringByUse. Configuring this phase as an alternative to the normal  AddressLoweringPhase requires modifying the hotspot suite providers so that common provision is done in a base phase HotSpotSuitesProviderBase. The CPU-specific suite provision that adds in the AddressLowering or AddressLoweringByUse phase is done in a subclass specialised for either AMD64/SPARC or for AArch64. More details in RFR which will be sent to graal dev list.


graal/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64AddressNode.java
```diff
@@ -76,7 +76,7 @@ public void generate(NodeLIRBuilderTool gen) {
         AllocatableValue indexReference;
         if (addressingMode.equals(AddressingMode.IMMEDIATE_UNSCALED)) {
             indexReference = LIRKind.derivedBaseFromValue(indexValue);
-            throw GraalError.unimplemented();
+            // throw GraalError.unimplemented();
```
dougxc: Please remove commented out code.

graal/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64AddressNode.java
```diff
@@ -120,12 +120,17 @@ public int getDisplacement() {
...
+    public void setDisplacement(int displacement, AArch64Address.AddressingMode addressingMode) {
         this.displacement = displacement;
+        this.addressingMode = addressingMode;
     }
 
     @Override
     public long getMaxConstantDisplacement() {
         return Long.MAX_VALUE;
     }
+    public AddressingMode getAddressingMode()
```
dougxc: Don't put opening `{`'s on a new line. Have you run `mx gate` locally? This should be caught by Checkstyle. You can run checkstyle standalone with `mx checkstyle`. You'll also probably want to run `mx eclipseformat` and `mx findbugs`.

dougxc:
Rather than introducing `HotSpotSuitesProviderBase`, I think it would be better to factor out:
```
    protected Phase createAddressLoweringPhase() {
        return new AddressLoweringPhase(addressLowering);
    }
```
which can be overridden by AArchHotSpotSuites.

adinn:
On 03/04/17 13:43, Douglas Simon wrote:
> Rather than introducing |HotSpotSuitesProviderBase|, I think it would be
> better to factor out:
> 
> |protected Phase createAddressLoweringPhase() { return new
> AddressLoweringPhase(addressLowering); } |
> 
> which can be overridden by AArchHotSpotSuites.

I considered doing something like that but I don't believe it is
possible to use overriding the way you are suggesting without also
adding a subclass for the AMD64/SPARC case. However, I think a slightly
different change to the one you suggest may avoid the need for any
overriding/subclassing of HotSpotSuitesProvider.

Perhaps I should explain each of those points in turn.

If you look at the changes I introduced you will notice that the
original constructor for HotSpotSuitesProvider accepted an
AddressLoweringPhase.AddressLowering. This was stored in a local
variable of HotSpotSuitesProvider and got used later by createSuites()
to parameterise the call to the constructor for AddressLoweringPhase
which creates the value passed to findPhase.add(). With your suggestion
above this field would be retained in class HotSpotSuitesProvider (your
suggested method, createAddressLoweringPhase, refers to this field).

The problem with this is that the AArch64HotspotSuitesProvider needs to
be passed an AddressLoweringByUsePhase.AddressLoweringByUse. Since, this
is a completely different interface it cannot be used to parameterise
the constructor call for HotspotSuitesProvider.

When factoring out the superclass, HotSpotSuitesProviderBase, I moved
this field down into the subclass (the new version of
HotSpotSuitesProvider). A value for this field is passed to the
subclass's constructor but it does not get passed in the super
constructor call. This is ok because the code which uses it is also now
located in the overridden version of createSuites. i.e. the parent class
is now independent of the lowering implementor.

This change allows AArch64HotspotSuitesProvider to accept an
implementation of AddressLoweringByUsePhase.AddressLoweringByUse which
is also only ever processed local to the subclass. It avoids having to
pass a redundant implementation of AddressLoweringPhase.AddressLowering
in the parent constructor call.

The change you are requesting -- overriding HotSpotSuitesProvider
without relocating the AddressLowering field into a non-AArch64 subclass
-- will not really work properly to achieve this decoupling.
AArch64HotspotSuitesProvider will still be required to pass an
AddressLowering implementation to the parent class even though it will
then be ignored.

What /will/ work is to change the constructor for HotSpotSuitesProvider
so that it accepts a Phase, either an AddressLoweringPhase instance or
an AddressLoweringByUsePhase. This is generic, unlike the current API
which expects a parameter to an architecture-specific phase constructor.
The relevant Phase instance can be constructed in the relevant arch's
BackendFactory when the HotSpotSuite{Provider is created.

Would you like me to implement that?

regards,


Andrew Dinn
-----------
Senior Principal Software Engineer
Red Hat UK Ltd
Registered in England and Wales under Company Registration No. 03798903
Directors: Michael Cunningham, Michael ("Mike") O'Neill, Eric Shander

adinn:
On 03/04/17 13:43, Douglas Simon wrote:
> *@dougxc* commented on this pull request.
> . . .
> Please remove commented out code.

Will do.

> ------------------------------------------------------------------------
> . . .
> Don't put opening |{|'s on a new line. Have you run |mx gate| locally?
> This should be caught by Checkstyle. You can run checkstyle standalone
> with |mx checkstyle|. You'll also probably want to run |mx
> eclipseformat| and |mx findbugs|.

Hmm, mx gate? mx checkstyle? ... right, the PR will be vetted accordingly!

regards,


Andrew Dinn
-----------
Senior Principal Software Engineer
Red Hat UK Ltd
Registered in England and Wales under Company Registration No. 03798903
Directors: Michael Cunningham, Michael ("Mike") O'Neill, Eric Shander

dougxc:
Wrt `AddressingLowering` refactoring, I overlooked the `HotSpotSuitesProvider.addressLowering` field. With that in mind, I still think there should be a abstract `Phase createAddressLowering()` method in `HotSpotSuitesProvider` and there will now be 3 subclasses of `HotSpotSuitesProvider`. The SPARC and AMD64 ones will be virtually identical. I don't think we want to start instantiating phases outside of `createSuites`.

adinn:
hi Doug,

Point taken regarding keeping the phase construction internal to the suites provider. I'll prepare another patch & PR where I have 3 subclasses implementing abstract method createAddressLowering.

graal/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/AddressLoweringByUsePhase.java
```diff
@@ -0,0 +1,104 @@
...
+import org.graalvm.compiler.nodes.memory.address.OffsetAddressNode;
+import org.graalvm.compiler.nodes.memory.address.RawAddressNode;
+import org.graalvm.compiler.nodes.util.GraphUtil;
+import org.graalvm.compiler.phases.Phase;
+/**
+ * Created by adinn on 21/03/17.
+ */
+public class AddressLoweringByUsePhase extends Phase
+{
+    public abstract static class AddressLoweringByUse
```
gilles-duboscq: This looks like it could be an interface

graal/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/AddressLoweringByUsePhase.java
```diff
@@ -0,0 +1,104 @@
...
+            // } else if (node instanceof PrefetchAllocateNode) {
+                // PrefetchAllocateNode prefetchAllocateNode = (PrefetchAllocateNode)node;
+                // AddressNode address = prefetchAllocateNode.getAddress();
+                // lowered = lowering.lower(prefetchAllocateNode, address.getBase(), address.getIndex());
+            } else {
+                continue;
+            }
+            if (lowered != address) {
+                address.replaceAtUsages(lowered);
+                GraphUtil.killWithUnusedFloatingInputs(address);
```
gilles-duboscq: So we do the lowering based on the first usage?
This means other usages will have be happy (correctness and performance-wise) with the shape chosen by the first usage.

gilles-duboscq:
@adinn you don't necessarily have to close this PR, you can iterate in this PR, it will be updated if it's open and you push to your branch.

adinn:
Hi Gilles,

On 03/04/17 16:25, Gilles Duboscq wrote:
> @adinn <https://github.com/adinn> you don't necessarily have to close
> this PR, you can iterate in this PR, it will be updated if it's open and
> you push to your branch.

I thought in view of the need to undo some class renaming a new PR might
be easier to manage.

If you think it is better to use the same branch then I can reopen it.

regards,


Andrew Dinn
-----------
Senior Principal Software Engineer
Red Hat UK Ltd
Registered in England and Wales under Company Registration No. 03798903
Directors: Michael Cunningham, Michael ("Mike") O'Neill, Eric Shander

gilles-duboscq:
it's fine for me either way

PR #268, opened 2017-04-12
Aarch64 backend patch2

adinn:
another small error in the emit code. this patch fixes an exception in CmpressPointer and UncompressPointer which gets thrown when the base register has value IllegalValue (which happens legitimately with Metadata pointers)


adinn:
closing because merged into earlier request

PR #272, opened 2017-04-27
Check JVMCI_VERSION_CHECK != ignore in mx

smarr:
The JVMCI version check isn't robust to strange version format strings.
When I compile the JVMCI locally from my git repo, it has a non-standard version string that leads to the following error:

```
$ ../mx/mx clean
Exception in thread "main" java.lang.NumberFormatException: For input string: "unknown-Artemis-5"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.graalvm.compiler.hotspot.JVMCIVersionCheck.check(JVMCIVersionCheck.java:82)
	at org.graalvm.compiler.hotspot.JVMCIVersionCheck.main(JVMCIVersionCheck.java:137)
```

@zapster suggested that the check shouldn't be done in the first place if `JVMCI_VERSION_CHECK=ignore` is set.
The attached patch avoids the check in that case.


dougxc:
I think we should contain handling of the JVMCI_VERSION_CHECK env var to the Java code. I'll make that change and fix the version number parsing at the same time. Thanks for reporting!

smarr:
Note, this doesn't seem to be a save change:

```
    if os.environ['JVMCI_VERSION_CHECK'] != 'ignore': 
  File "/home/travis/build/graalvm/graal-core/venv/lib/python2.7/UserDict.py", line 23, in __getitem__
    raise KeyError(key)
KeyError: 'JVMCI_VERSION_CHECK'
```

So, I guess, something like `.get('JVMCI_VERSION_CHECK', None)` is needed.

smarr:
Solved with https://github.com/graalvm/graal/commit/52b609455518a461c317d6a6bbe620ad2e5de897

Thanks

PR #275, opened 2017-05-18
Aarch64 address lowering2

adinn:
Here is a pull request that includes a complete AArch64 lowering by use phase. This takes into account the usage for an address so as to allow SCALED_IMMEDIATE addressing mode to be used (as well as also using SCALED_IMMEDIATE where appropriate). This removes the idiom where smneed to load small constants into registers


tkrodriguez:
Can you bring these changes forward onto the new merged repo in https://github.com/graalvm/graal?  There are a bunch of renames you'll have to adjust for.  Also there appear to be some formatting problems.  Can you also resolve any issues reported by mx checkstyle?

adinn:
On 23/05/17 07:32, Tom Rodriguez wrote:
> Can you bring these changes forward onto the new merged repo in
> https://github.com/graalvm/graal? There are a bunch of renames you'll
> have to adjust for.

I have done so and will post a new PR as soon as I have verified correct
operation.

Unfortunately, after re-applying what are essentially the same changes I
have now encountered some further unexpected exceptions during testing
(so far only when running netbeans rather than in a simple reproducer).
At the moment my belief is that these relate to ByteBuffer accesses --
an index pulled out of a ByteBuffer and then passed toArrayList.get is
causing an IllegalArgumentException because it is way off the scale.

I am still investigating the cause but this looks like another subtle
interaction with something in an unrelated part of the code (like the
use of an invalid address type I corrected i the last version). IN
consequence, it may well take some time to isolate an diagnose the
source of the problem.

>                      Also there appear to be some formatting problems.
> Can you also resolve any issues reported by mx checkstyle?

I will do so before submitting the next PR.

regards,


Andrew Dinn
-----------
Senior Principal Software Engineer
Red Hat UK Ltd
Registered in England and Wales under Company Registration No. 03798903
Directors: Michael Cunningham, Michael ("Mike") O'Neill, Eric Shander

tkrodriguez:
Thanks.  I did manage to update your repo locally and run it through our internal gate.  Apart from the checkstyle stuff it all worked fine which is good news.  checkstyle didn't like your copyright additions so we'll need to fix that somehow.  I'll figure out what to do about that.

tkrodriguez:
I'm pushing changes to checkstyle that will accept your copyright additions, though they need to come after the Oracle one, so just swap those lines.  checkstyle isn't that flexible so it was easier to allow arbitrary copyrights after the oracle one instead of permitting arbitrary orderings.

adinn:
Thanks for inserting the copyright fix. I'll ensure the Red Hat copyright line follows the Oracle one in the next PR I post.

tkrodriguez:
The copyright problem should be fixed by fcc9a39

